PR #5, opened 2022-04-12
Redirect standard logs

amckenzie132:
If the log level > INFO, redirect log messages created by the standard library's global logger to the configured proxy logger with `component=standard` and level `DEBUG`

If the log level <= `INFO`, then disable standard logging.


alexh-sauce:
Is there any way we can trigger a write to stdout in a test to capture the output?

waggledans (approved):

alexh-sauce (approved):
LGTM

PR #14, opened 2022-06-29
SC-3812 Add basic auth

alexh-sauce:
Add `siteCredentials` to the forwarder.

This will allow the forwarder to pass credentials to the requested site without the original requester needing any credentials.

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -16,6 +16,8 @@ var (
 	localProxyURI    string
 	upstreamProxyURI string
 
+	siteCredentials []string
+
 	pacProxiesCredentials []string
 	pacURI                string
 
@@ -46,10 +48,10 @@ Note: Can't setup upstream, and PAC at the same time.
 
   Start a proxy listening to http://0.0.0.0:8085:
   $ forwarder run -l "http://0.0.0.0:8085"
-  
+
   Start a protected proxy:
   $ forwarder run -l "http://user:pwd@localhost:8085"
-  
+
   Start a protected proxy, forwarding connection to an upstream proxy running at
   http://localhost:8089:
   $ forwarder run \
@@ -61,19 +63,19 @@ Note: Can't setup upstream, and PAC at the same time.
   $ forwarder run \
     -l "http://user:pwd@localhost:8085" \
     -u "http://user1:pwd1@localhost:8089"
-	
+
   Start a protected proxy, forwarding connection to an upstream proxy, setup via
   PAC - server running at http://localhost:8090:
   $ forwarder run \
     -l "http://user:pwd@localhost:8085" \
     -p "http://localhost:8090"
-	
+
   Start a protected proxy, forwarding connection to an upstream proxy, setup via
   PAC - protected server running at http://user2:pwd2@localhost:8090:
   $ forwarder run \
     -l "http://user:pwd@localhost:8085" \
     -p "http://user2:pwd2@localhost:8090"
-	
+
   Start a protected proxy, forwarding connection to an upstream proxy, setup via
   PAC - protected server running at http://user2:pwd2@localhost:8090, specifying
   credential for protected proxies specified in PAC:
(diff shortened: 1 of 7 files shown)


amckenzie132 (approved):
LGTM

cmd/run.go
```diff
@@ -120,6 +122,7 @@ func init() {
 	runCmd.Flags().StringSliceVarP(&dnsURIs, "dns-uri", "n", nil, "sets dns URI")
 	runCmd.Flags().StringVarP(&pacURI, "pac-uri", "p", "", "sets URI to PAC content, or directly, the PAC content")
 	runCmd.Flags().StringSliceVarP(&pacProxiesCredentials, "pac-proxies-credentials", "d", nil, "sets PAC proxies credentials using standard URI format")
+	runCmd.Flags().StringSliceVarP(&siteCredentials, "site-credentials", "d", nil, "sets site based credentials")
```
waggledans, line 125: typo here, `-d` is already used. I'd recommend using `StringSliceVar`, it's not a common option, no need to add a shortcut.

In general, I don't like passing credentials in this form: `host:port:user:pass`. I'd recommend using a more conventional `user:pass@url` (with the idea in mind that users of `forwarder` as a module would translate `host:port:user:pass` into `user:pass@url` 🤷‍♂️ . It's optional though, only if you agree.

waggledans:
actually, committed some minor code refactoring if you don't mind (it's just easier than commenting), feel free to discard it if you don't like it, it's your MR ;) @alexh-sauce

waggledans:
I'd also suggest benchmarking. I added `make bench` just to check the baseline but you may want to try e2e benchmarking with `vegetta` to get a sense of the perf impact.

cmd/run.go
alexh-sauce: Oh yeah, that's a great idea. I was annoyed that I couldn't reuse the existing PAC auth code.

alexh-sauce:
Did some benchmarks, and the performance looks the same on the non-auth path.

```
$ cat master1.bench 
goos: linux
goarch: amd64
pkg: github.com/saucelabs/forwarder/pkg/proxy
cpu: AMD A8-7600 Radeon R7, 10 Compute Cores 4C+6G  
BenchmarkNew-4              1347            769087 ns/op
PASS
ok      github.com/saucelabs/forwarder/pkg/proxy        9.992s
```


```
$ cat auth.bench
goos: linux
goarch: amd64
pkg: github.com/saucelabs/forwarder/pkg/proxy
cpu: AMD A8-7600 Radeon R7, 10 Compute Cores 4C+6G  
BenchmarkNew-4              1417            769744 ns/op
PASS
ok      github.com/saucelabs/forwarder/pkg/proxy        9.988s
```

pkg/proxy/proxy.go
```diff
@@ -364,39 +373,85 @@ func setupPACUpstreamProxyConnection(p *Proxy, ctx *goproxy.ProxyCtx) error {
...
 		if err != nil {
-			return nil, err
+			return nil, nil, nil, "", err
 		}
 
 		// Get the base64 of the credentials
 		pass, found := uri.User.Password()
 		if !found {
-			return nil, fmt.Errorf("password not found in %s", credentialText)
+			return nil, nil, nil, "", fmt.Errorf("password not found in %s", credentialText)
```
waggledans, line 403: I'd rewrite that.

```
var ErrPasswordNotFound = customerror.NewInvalidError("password not found")
...
fmt.Errorf("%w: %s", ErrPasswordNotFound, credentialText)
```

waggledans:
👋  Im going through the code jetlagged so I apologize if my comments don't make any sense. I think issues with your change are:

- Proxy struct becomes unnecessary complex due to a marginal use of site creds
    - 	siteCredentials
    - siteCredentialsHost
    - siteCredentialsPort
    - siteCredentialsWildcard
- OnRequest handler does several unrelated things (ie `maybeAddAuthHeader` is called all the time, I don't like it)
- `proxy.New` always includes `siteCreds` param that is rarely used (I'd move it to `Options`)
- Some other minor things.

I think I'll try to implement the changes and submit them as a separate MR. Will do it soon.

waggledans:
> I think I'll try to implement the changes and submit them as a separate MR. Will do it soon.

here it is: https://github.com/saucelabs/forwarder/pull/15

pkg/proxy/proxy.go
waggledans: fixed it

waggledans (approved):

amckenzie132 (approved):

PR #15, opened 2022-07-11
Refactoring site credentials handling

waggledans:
Refactoring the implementation of the "site credentials" feature in https://github.com/saucelabs/forwarder/pull/14

Improvements are:

- site credentials are handled in a separate callback that is added ONLY if AT LEAST one site is configured. It's a special use case and it shouldn't affect the main flow
- support schema in the credentials URL (just a convenience)
- support setting credentials via an env var: `FORWARDER_SITE_CREDENTIALS`
- refactor the site credentials matcher out of the main `Proxy` struct to keep the implementation somewhat cleaner
- Also moving a `siteCredentials` param to `Options` (to keep the `New` API cleaner)


amckenzie132 (commented):
LGTM

Issue #22, opened 2022-09-08
Run linter on the internal packages


mmatczuk:
One possible solutions is to move the `internal` packages up or even unpack some of them to project root.

PR #43, opened 2022-09-12
Binary building modernized

mmatczuk:
The motivation for this work is improving package structure, and running `golangci-lint` on all the code see #22.

This patchset focuses on moving `cmd` to `cmd/forwarder` and removing `main.go` from git root.
At the moment the package `github.com/saucelabs/forwarder` is in fact `main`, this blocks #22. 

This work also...
- Removes air
- Fixes GoReleaser configuration
- Further cleans makefile

Please find details in commit messages.


mmatczuk:
Congratulations to self, build fails with `Error return value of `rootCmd.Execute` is not checked (errcheck)`, first results come faster than I anticipated.

Makefile
```diff
@@ -9,23 +9,21 @@ export PATH  := $(GOBIN):$(PATH)
 
 .PHONY: install-dependencies
 install-dependencies:
-	@rm -Rf bin
-	go install github.com/cosmtrek/air@latest
```
waggledans, line 13: I, personally, don't use air, it doesn't seem to be helpful to me to see my changes as they happen so I don't have any opinion on that one

waggledans (approved):

PR #66, opened 2022-09-20
Configuration management improvements

### LINKED ISSUE #50: Proxy constructor shall not evaluate env variables
Users should be able to configure the binary with command line flags and env variables.
This can be easily automated with Cobra + Viper, see [this issue](https://github.com/spf13/viper/issues/85).
All of that shall happen in `cmd/forwarder`, Proxy gets configuration it can validate...

### LINKED ISSUE #55: validation: error messages are useless
Example 

```
'ProxyConfig.LocalProxyURI' Error:Field validation for 'LocalProxyURI' failed on the 'proxyURI' tag
```

### LINKED ISSUE #61: Allow importing the run command in 3rd party packages
Move cmd/forwarder/run.go and cmd/forwarder/run package and allow to import the forwarder proxy cobra command.
Consider renaming run to proxy.

### LINKED ISSUE #67: validation: DNS addresses MUST NOT be host name
Consider this

```
// dial makes a new connection to the provided server (which must be
// an IP address) with the provided network type, using either r.Dial
// (if both r and r.Dial are non-nil) or else Dialer.DialContext.
func (r *Resolver) dial(ctx context.Context, network, server string) (Conn, error) {
	// Calling Dial here is scary -- we have to be sure not to
	// dial a name that will require a DNS lookup, or Dial will
	// call back here to translate it. The DNS config parser has
	// already checked that all the cfg.servers are IP
	// addresses, which Dial will use without a DNS lookup.
	var c Conn
	var err error
	if r != nil && r.Dial != nil {
		c, err = r.Dial(ctx, network, server)
	} else {
		var d Dialer
		c, err = d.DialContext(ctx, network, server)
	}
```

Since we replace dial function in Resolver we must make sure that we would not end up doing a name resolution.

### LINKED ISSUE #68: Add general support for env vars
At the moment support for env vars is limited to selected few that are evaluated in Proxy constructor.

mmatczuk:
This patchset solves many intertwined issues of configuration parsing, passing, and validation.

* #55 
* #67 and others like demanding that DNS port must be get 53
* #50
* #61
* #68

# CLI

## Env vars

Each CLI flag has a documented env var that can be used instead of passing the flag directly.

```
Flags:
  -n, --dns-uri                           sets dns URI (env FORWARDER_DNS_URI) (default [])
  -h, --help                              help for run
      --local-proxy-auth                  sets local proxy basic auth in the form of username:password (env FORWARDER_LOCAL_PROXY_AUTH)
  -l, --local-proxy-uri                   sets local proxy URI (env FORWARDER_LOCAL_PROXY_URI) (default http://localhost:8080)
      --log-file-level string             sets the log file level (env FORWARDER_LOG_FILE_LEVEL) (default "info")
      --log-file-path string              sets the log file path (env FORWARDER_LOG_FILE_PATH) (default "/var/folders/7q/nyynjpwj5p19npby48ykjpx00000gn/T/proxy-2022-09-23T14-34-14.log")
      --log-level string                  sets the log level (env FORWARDER_LOG_LEVEL) (default "info")
  -d, --pac-proxies-credentials strings   sets PAC proxies credentials using standard URI format (env FORWARDER_PAC_PROXIES_CREDENTIALS)
  -p, --pac-uri                           sets URI to PAC content, or directly, the PAC content (env FORWARDER_PAC_URI)
  -t, --proxy-localhost                   if set, will proxy localhost requests to an upstream proxy - if any (env FORWARDER_PROXY_LOCALHOST)
      --site-credentials strings          sets target site credentials (env FORWARDER_SITE_CREDENTIALS)
      --upstream-proxy-auth               sets upstream proxy basic auth in the form of username:password (env FORWARDER_UPSTREAM_PROXY_AUTH)
  -u, --upstream-proxy-uri                sets upstream proxy URI (env FORWARDER_UPSTREAM_PROXY_URI)
```

This is implemented by integrating Viper with Cobra, order of evaluation is Defaults, Env, Flags.

## Validation

Validation of flags produces meaningful error messages, examples:

```
go run ./cmd/forwarder/ run --local-proxy-uri http://foo:29
...
invalid argument "http://foo:29" for "-l, --local-proxy-uri" flag: invalid hostname: foo is too short
```

```
go run ./cmd/forwarder/ run --dns-uri 1.1.1.1 --dns-uri udp://google.com:53
...
invalid argument "udp://google.com:53" for "-n, --dns-uri" flag: invalid hostname: google.com DNS must be an IP address
```

The validation library is removed, validation rules are revised, improved, and tested. 

## Subcomand packages

Each subcommand is put to a separate package for better separation of concerns.
It allows to embed the `run` command in a different binary as a subcommand without any additional work!  

# ProxyConfig

ProxyConfig holds URIs as *url.URL. In tests and `sc` code we always work with URL instances and only then encode them. Now this can be avoided. In addition to that `ProxyConfig::Validate` function shares code with validations used in Cobra. 

## Cobra flags bindings

Normally (in the past) binding a custom type to Cobra flag required adding a new type and implementing `pflag.Value` and/or `pflag.SliceValue` for slices. This is purely boilerplate code that still needs o be tested. I addressed that by implementing [anyflag](https://github.com/mmatczuk/anyflag) library. It implements the interfaces with Go generics, callers must only provide the parser function. It is used in this PR to streamline the binding. 

# Proxy

The parsed urls fields are removed, some additional renames take place.

## DNS

DNS code is streamlined and put to separate file. 

Fixes #50 
Fixes #55 
Fixes #61 
Fixes #67
Fixes #68


amckenzie132:
Looks great
Small change: in a couple comments `upd` -> `udp`

amckenzie132 (approved):

mmatczuk:
Thanks @amckenzie132! 

Fixed `s/upd/upd` in `proxyconfig.go`.
Updated anyflag.

PR #78, opened 2022-09-28
proxy: fix initialization

### LINKED ISSUE #29: setupHandlers is odd and buggy
The [setupProxyHandlers](https://github.com/saucelabs/forwarder/blob/bb67daa2958e88d2c784bbddb1c0a1ccfdaebb28/pkg/proxy/proxy.go#L625) function that configures goproxy uses a strange method where everything is always configured on per request basis.

It looks like it creates race condition... 
This code called for every request clearly changes the server state.

```go
// setupUpstreamProxyConnection forwards connections to an upstream proxy.
func setupUpstreamProxyConnection(ctx *goproxy.ProxyCtx, uri *url.URL) {
	ctx.Proxy.Tr.Proxy = http.ProxyURL(uri)
```

Also, [HandleConnectFunc](https://github.com/saucelabs/forwarder/blob/bb67daa2958e88d2c784bbddb1c0a1ccfdaebb28/pkg/proxy/proxy.go#L626) returns a nil ConnectAction while it should [accept the connect](https://github.com/elazarl/goproxy/blob/fbd10ff4f5a16de73dca5030fc12245548f76141/https.go#L126) - this code path seems not to be tested.

Normally we should be using `ReqCondition` like

```
// Typical usage:
//	proxy.OnRequest(UrlIs("example.com/foo"),UrlMatches(regexp.MustParse(`.*\.exampl.\com\./.*`)).Do(...)
```

those can be also custom interface implementations that allow for more dynamic bhv.

### LINKED ISSUE #31: ExampleNew does not proxy anything
The promise 

```
// client -> protected local proxy -> protected pac server - connection setup -> protected upstream proxy -> protected target.
```

The reality

```
component=proxy output=console level=trace timestamp=2022-09-08T14:17:09+02:00 message=Not proxifying request to localhost URL: http://127.0.0.1:62270/
```

It's easy to fix by adding `ProxyLocalhost: true`.

### LINKED ISSUE #74: proxy: set DNS first
DNS is currently [configured after PAC](https://github.com/saucelabs/forwarder/blob/43c34325af1d724d5dcfe97e168f96a0f4c4ad96/proxy.go#L71-L78).

This results in PAC server address being resolved with stock DNS server not the DNS provided to the command.

### LINKED ISSUE #76: proxy: localhost proxying exposes proxy local services
When `ProxyLocalhost` is disabled, user can call services listing on localhost on the proxy machine.
This is exactly what we can see in #31 logs, local proxy does not hit upstream, instead it uses its own localhost to deliver the result.

Related to #31

mmatczuk:
This patchset reimplements goproxy setup so that the proxy is not reconfigured in runtime.
In addition to that it fixes missing / wrong configurations. See commit messages for details.
 
Fixes #29
Fixes #31
Fixes #74 
Fixes #76

### DIFF
--- a/proxy.go
+++ b/proxy.go
@@ -68,6 +68,8 @@ func NewProxy(cfg *ProxyConfig, log Logger) (*Proxy, error) {
 		proxy:    goproxy.NewProxyHttpServer(),
 		log:      log,
 	}
+	p.setupDNS()
+
 	if p.config.PACURI != nil {
 		pacParser, err := pacman.New(p.config.PACURI.String(), p.config.PACProxiesCredentials...)
 		if err != nil {
@@ -75,7 +77,7 @@ func NewProxy(cfg *ProxyConfig, log Logger) (*Proxy, error) {
 		}
 		p.pacParser = pacParser
 	}
-	p.setupDNS()
+
 	p.setupProxy()
 
 	return p, nil
--- a/proxy.go
+++ b/proxy.go
@@ -100,7 +100,6 @@ func (p *Proxy) setupProxy() {
 	// See: https://github.com/golang/go/issues/28866
 	// See: https://github.com/elazarl/goproxy/issues/306
 	p.proxy.KeepHeader = true
-	p.proxy.Tr = &http.Transport{}
 
 	p.setupBasicAuth()
 	p.setupLocalhostProxy()
--- a/proxy.go
+++ b/proxy.go
@@ -65,7 +65,6 @@ func NewProxy(cfg *ProxyConfig, log Logger) (*Proxy, error) {
 	p := &Proxy{
 		config:   *cfg,
 		userInfo: m,
-		proxy:    goproxy.NewProxyHttpServer(),
 		log:      log,
 	}
 	p.setupDNS()
@@ -91,6 +90,7 @@ func (p *Proxy) setupDNS() {
 }
 
 func (p *Proxy) setupProxy() {
+	p.proxy = goproxy.NewProxyHttpServer()
 	p.proxy.Logger = goproxyLogger{p.log}
 	p.proxy.Verbose = true
 	p.proxy.KeepDestinationHeaders = true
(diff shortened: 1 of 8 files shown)


mmatczuk:
This PR also contains some work regarding a larger fix the init process as described in #29.
It can be moved to separate PRs if needed.
I think it makes sense to keep them here for simplicity.

proxy.go
```diff
@@ -98,246 +94,150 @@ func (p *Proxy) setupProxy() {
...
-		req.URL.String(),
-	)
+	p.setupProxyBasicAuth()
+	p.setupSiteBasicAuth()
 }
 
-// Removes any upstream proxy settings.
-func resetUpstreamSettings(ctx *goproxy.ProxyCtx) {
-	ctx.Proxy.ConnectDial = nil
+func (p *Proxy) setupLocalhostProxy() {
```
alexh-sauce, line 115: I think this changed the logic for `localhost` handling. It's possible the original code was wrong as well. :)

The intention is to only pass requests for `localhost` to a parent proxy (set via upstream or PAC) if specifically enabled. 

However, if *disabled*, requests coming in to `localhost` should be sent to localhost directly regardless of the parent proxy settings. If I'm reading this correctly, if `p.config.ProxyLocalhost == false` and a request came in, it would return a 502 error.

If there is no parent proxy set, it would always go to `localhost` regardless of `p.config.ProxyLocalhost`.

alexh-sauce (changes requested):

proxy.go
mmatczuk: Precisely. Let's take a look at it case by case:

OLD

|Mode|ProxyLocalhost=true|ProxyLocalhost=false|
|--|--|--|
|DIRECT       | Dial localhost  | Dial localhost |
|UPSTREAM| Proxy localhost | Dial localhost |

NEW

|Mode|ProxyLocalhost=true|ProxyLocalhost=false|
|--|--|--|
|DIRECT       | Dial localhost  | Bad GW |
|UPSTREAM| Proxy localhost | Bad GW |

The original behavior has two problems
1. It's inconsistent
1. It is security threat, you think the localhost is not exposed but it is

PR #103, opened 2022-10-21
streamline logging configuration

mmatczuk:


config.go
```diff
@@ -146,3 +148,22 @@ func isPort(port string) bool {
 
 	return p >= 1 && p <= 65535
 }
+
+// OpenFileParser returns a parser tha calls os.OpenFile.
```
alexh-sauce, line 152: s/tha/that

mmatczuk: Done.

PR #135, opened 2022-11-24
e2e: port tests to httpexpect

mmatczuk:
e2e is now a separate go module independent of forwarder. This is done to avoid mixing dependencies.
Tests are rewritten and improved.

Default dev setup is changed from HTTPS to HTTP.
To enable working with the dev env it's now sufficient to set DEV environment variable.

All framework tools are put to the framework.go file.


mmatczuk:
tx

Issue #129, opened 2022-11-21
martian: add more fine grained timeouts


mmatczuk:
The `http.Server` has a bunch of timeouts that can be configured.

```
	// ReadTimeout is the maximum duration for reading the entire
	// request, including the body. A zero or negative value means
	// there will be no timeout.
	//
	// Because ReadTimeout does not let Handlers make per-request
	// decisions on each request body's acceptable deadline or
	// upload rate, most users will prefer to use
	// ReadHeaderTimeout. It is valid to use them both.
	ReadTimeout time.Duration

	// ReadHeaderTimeout is the amount of time allowed to read
	// request headers. The connection's read deadline is reset
	// after reading the headers and the Handler can decide what
	// is considered too slow for the body. If ReadHeaderTimeout
	// is zero, the value of ReadTimeout is used. If both are
	// zero, there is no timeout.
	ReadHeaderTimeout time.Duration

	// WriteTimeout is the maximum duration before timing out
	// writes of the response. It is reset whenever a new
	// request's header is read. Like ReadTimeout, it does not
	// let Handlers make decisions on a per-request basis.
	// A zero or negative value means there will be no timeout.
	WriteTimeout time.Duration

	// IdleTimeout is the maximum amount of time to wait for the
	// next request when keep-alives are enabled. If IdleTimeout
	// is zero, the value of ReadTimeout is used. If both are
	// zero, there is no timeout.
	IdleTimeout time.Duration
```

In addition to that `TLSHandshakeTimeout` should be added.

Now we have one fits all `Timeout` that's being use as `ReadHeaderTimeout`, `WriteTimeout` and `TLSHandshakeTimeout`. 
For plain HTTP requests that means that whole request response cycle needs to be completed under the t/o otherwise we cannot write the response.

PR #157, opened 2022-12-16
http_server: add read and write timeouts

Choraden:
This patch integrates forwarder with timeout changes introduced in https://github.com/saucelabs/martian/pull/1.

### DIFF
--- a/http_server.go
+++ b/http_server.go
@@ -66,12 +66,14 @@ func h2TLSConfigTemplate() *tls.Config {
 }
 
 type HTTPServerConfig struct {
-	Protocol        Scheme        `json:"protocol"`
-	Addr            string        `json:"addr"`
-	CertFile        string        `json:"cert_file"`
-	KeyFile         string        `json:"key_file"`
-	ReadTimeout     time.Duration `json:"read_timeout"`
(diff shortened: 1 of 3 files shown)


http_server.go
```diff
@@ -80,9 +82,10 @@ type HTTPServerConfig struct {
 
 func DefaultHTTPServerConfig() *HTTPServerConfig {
 	return &HTTPServerConfig{
-		Protocol:    HTTPScheme,
-		Addr:        ":8080",
-		ReadTimeout: 5 * time.Second,
+		Protocol:     HTTPScheme,
+		Addr:         ":8080",
+		ReadTimeout:  5 * time.Second,
```
mmatczuk, line 87: This is wrong, we should be using read header with some high value like 1m by default.

mmatczuk:
Also note the `ReadTimeout: 0, // TODO(mmatczuk): we need a more fine-grained timeouts see #129`.
There are two distinct modes here, proxy and server, they have different requirements.

mmatczuk:
FYI we still have `hp.proxy.SetTimeout(hp.config.ReadTimeout)`

mmatczuk (approved):

PR #162, opened 2022-12-20
config: add support for port '*'

### LINKED ISSUE #57: userInfoMatcher add support for port '*'
Related work https://github.com/saucelabs/libsauceconnect/commit/d9c5a5436d9378f5417524ff59b16391976296e9

Choraden:
Parsing user:password@host:port string will check if port is '*' and change it to '0'.

Fixes #57

### DIFF
--- a/config.go
+++ b/config.go
@@ -43,10 +43,18 @@ func validatedUserInfo(ui *url.Userinfo) error {
 	return nil
 }
 
+func wildcardPortTo0(val string) string {
+	s := strings.Split(val, ":")
+	if s[len(s)-1] == "*" {
+		s[len(s)-1] = "0"
+	}
+	return strings.Join(s, ":")
+}
+
 // ParseHostPortUser parses a user:password@host:port string into HostUser.
 // User and password cannot be empty.
 func ParseHostPortUser(val string) (*HostPortUser, error) {
-	u, err := url.Parse("http://" + val)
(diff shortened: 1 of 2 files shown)


config.go
```diff
@@ -51,9 +51,14 @@ func ParseHostPortUser(val string) (*HostPortUser, error) {
 		return nil, err
 	}
 
+	p := u.Port()
```
mmatczuk, line 54: Isn't adding it to CredentialsManager nicer and simpler?

mmatczuk: Also, a test would be good, CredentialsManager has the tests to add this

config.go
```diff
@@ -43,10 +43,18 @@ func validatedUserInfo(ui *url.Userinfo) error {
...
+		s[len(s)-1] = "0"
+	}
+	return strings.Join(s, ":")
+}
+
 // ParseHostPortUser parses a user:password@host:port string into HostUser.
 // User and password cannot be empty.
 func ParseHostPortUser(val string) (*HostPortUser, error) {
-	u, err := url.Parse("http://" + val)
+	u, err := url.Parse("http://" + wildcardPortTo0(val))
```
Choraden, line 57: I'm afraid, we can't move it to credentials matcher. 
url.Parse fails on port '*'.

mmatczuk: Gotcha.

PR #164, opened 2022-12-22
cmd/forwarder: group flags in help view

### LINKED ISSUE #136: cmd help group flags by prefix
There is quite a bit of flags. It would be much easier to digest them if we put the core flags at the beginning and then the
- api flags
- http flags
- log flags

In other words take out the core flags from between the other flags.

Choraden:
The flags are groupped:
- core
- api
- http
- log

Fixes #136


mmatczuk:
I think they do not have all the components from the list

```
- core
- api
- http
- log
```

PR #167, opened 2022-12-22
http_proxy: add host file watcher 

### LINKED ISSUE #123: Watch /etc/hosts and reload in memory representation
In the unlikely event of someone modifying the /etc/hosts we need to be able to reload it in runtime to avoid exposing localhost.

Choraden:
This patch removes reading and storing hosts from /etc/hosts.
Instead the proxy will use local resolver to lookup hosts.

 
Fixes #123


mmatczuk:
I have an alternative solution to the whole problem. It's inspired by [this post](https://stackoverflow.com/a/51839787).
We'd try to use all the file level ops and caching provided by DNS resolver without ever dialing to a DNS server.

The idea:
* Implement everything in resolver, we already have a custom [Dial() function](https://github.com/saucelabs/forwarder/blob/ee2846fe0fbff25d9c86b45c849ef08776b29431/dns.go#L80)
* Use special context parameter to taint the context
* In Dial if context is tainted return a special error that isLocalhost can understand.

Resolver needs to be extended so that it can work with empty Servers list.

http_proxy.go
```diff
@@ -266,27 +259,21 @@ func (hp *HTTPProxy) denyLocalhost() martian.RequestModifier {
...
-	// IP addresses.
-	if ip := net.ParseIP(h); ip != nil {
-		return ip.IsLoopback()
-	}
-	// IPv6 without a port number - Hostname returns the invalid value.
-	if ip := net.ParseIP(req.URL.Host); ip != nil {
-		return ip.IsLoopback()
+
+	if addrs, err := localhostResolver.LookupHost(context.Background(), h); err == nil {
+		// Take the first address from list.
```
mmatczuk, line 273: Why is this comment relevant?

http_proxy.go
```diff
@@ -266,27 +259,21 @@ func (hp *HTTPProxy) denyLocalhost() martian.RequestModifier {
 	}, errors.New("localhost access denied"))
 }
 
-// The goproxy.IsLocalHost is broken.
-// In addition, it doesn't work with hostnames that are not in /etc/hosts.
-// See https://github.com/elazarl/goproxy/issues/487
+var localhostResolver = nopResolver() //nolint:gochecknoglobals // This is local resolver.
```
mmatczuk, line 262: This is worth a comment to explaining what is it doing and why.

PR #173, opened 2022-12-29
modifiers: add request modifier that removes headers by prefix

### LINKED ISSUE #168: util: add request modifier that removes headers by prefix

Choraden:
Fixes #168

### DIFF
--- a/bind/flag.go
+++ b/bind/flag.go
@@ -40,6 +40,8 @@ func HTTPProxyConfig(fs *pflag.FlagSet, cfg *forwarder.HTTPProxyConfig) {
 	}
 	fs.VarP(anyflag.NewValue[forwarder.ProxyLocalhostMode](cfg.ProxyLocalhost, &cfg.ProxyLocalhost, anyflag.EnumParser[forwarder.ProxyLocalhostMode](proxyLocalhostValues...)),
 		"proxy-localhost", "t", "accept or deny requests to localhost, one of deny, allow, direct; in direct mode localhost requests are not sent to upstream proxy if present")
+
+	fs.StringSliceVar(&cfg.RemoveHeaders, "remove-headers", cfg.RemoveHeaders, "removes request headers if prefixes match (can be specified multiple times)")
 }
 
 func HTTPTransportConfig(fs *pflag.FlagSet, cfg *forwarder.HTTPTransportConfig) {
--- a/http_proxy.go
+++ b/http_proxy.go
@@ -12,6 +12,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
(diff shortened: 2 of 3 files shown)


Choraden:
Prefix is now formatted in canonic way. 

I've also done some research on this. 

```go
// Add adds the key, value pair to the header.
// It appends to any existing values associated with key.
func (h MIMEHeader) Add(key, value string) {
	key = CanonicalMIMEHeaderKey(key)
	h[key] = append(h[key], value)
}

// Del deletes the values associated with key.
func (h MIMEHeader) Del(key string) {
	delete(h, CanonicalMIMEHeaderKey(key))
}
```

Methods from http package auto translate keys to canonic way. Using this methods is safe. The only way one can break it is to create header in place: 
```go 
http.Header{
    "non canonical": nil,
},
```
But that would be a really stupid user. 

As long as forwarder is using http.ReadRequest we're good. 

@mmatczuk

mmatczuk:
Package shall be singular. Anyway, let's I'd like to suggest a few changes:
* Put that to http_proxy.go
* Do not export
* Rename to something more flexible ex. headerRemover
* Add this feature to proxy config so that it's usable from CLI

bind/flag.go
```diff
@@ -34,6 +34,7 @@ func HTTPProxyConfig(fs *pflag.FlagSet, cfg *forwarder.HTTPProxyConfig) {
 		"upstream-proxy", "u", "upstream proxy URL")
 	fs.BoolVarP(&cfg.ProxyLocalhost, "proxy-localhost", "t", cfg.ProxyLocalhost,
 		"proxy localhost requests to an upstream proxy")
+	fs.StringSliceVar(&cfg.HeaderRemover, "header-remover", cfg.HeaderRemover, "removes request headers by prefix (can be specified multiple times)")
```
mmatczuk, line 37: can we have a sane flag name for that?

mmatczuk: Shorter plese.

Choraden: `remove-headers` sounds nice to me

http_proxy.go
```diff
@@ -65,6 +66,7 @@ type HTTPProxyConfig struct {
 	RequestModifiers  []martian.RequestModifier  `json:"-"`
 	ResponseModifiers []martian.ResponseModifier `json:"-"`
 	CloseAfterReply   bool                       `json:"close_after_reply"`
+	HeaderRemover     []string                   `json:"header_remover"`
```
mmatczuk, line 69: ?

Choraden: changed to flag name i.e. `RemoveHeaders`

PR #179, opened 2023-01-10
http_proxy: add ProxyLocalhost=direct mode

### LINKED ISSUE #176: proxy: allow to directly handle localhost requests
This is a requirement mentioned in [this comment](https://github.com/saucelabs/forwarder/issues/76#issuecomment-1259661325). We need a way to express the need not to proxy localhost requests upstream.

mmatczuk:
ProxyLocalhost is changed to enum, available values are:
- deny (former false)
- allow (former true)
- direct

In the direct mode requests to localhost are allowed but are not sent to upstream proxy.

Fixes #176


Choraden (approved):
lgtm

PR #181, opened 2023-01-12
httplog: add new logger

Choraden:
This patch adds new http logging modes

### DIFF
--- a/httplog/logger.go
+++ b/httplog/logger.go
@@ -1,0 +1,187 @@
+// Copyright 2022 The forwarder Authors. All rights reserved.
+// Use of this source code is governed by a MPL
+// license that can be found in the LICENSE file.
+
+package httplog
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+
+	"github.com/google/martian/v3/messageview"
+	"github.com/saucelabs/forwarder/middleware"
+)
+
+type LoggerMode string
+
+const (
+	URLLogMode     LoggerMode = "url"
+	HeadersLogMode LoggerMode = "headers"
+	BodyLogMode    LoggerMode = "body"
+	ErrOnlyLogMode LoggerMode = "error"
+)
+
+func (m LoggerMode) Validate() error {
+	switch m {
+	case URLLogMode, HeadersLogMode, BodyLogMode, ErrOnlyLogMode:
+		return nil
+	}
+
+	return fmt.Errorf("log mode %s not found", m)
+}
+
+func ParseMode(val string) (LoggerMode, error) {
+	mode := LoggerMode(val)
+	switch mode {
+	case URLLogMode, HeadersLogMode, BodyLogMode, ErrOnlyLogMode:
+		return mode, nil
+	}
+
+	return "", fmt.Errorf("log mode %s not found", mode)
+}
+
+type Logger struct {
+	log  func(format string, args ...interface{})
+	mode LoggerMode
+}
+
+// NewLogger returns a logger that logs HTTP requests and responses.
+func NewLogger(logFunc func(format string, args ...interface{}), mode LoggerMode) *Logger {
+	return &Logger{
+		log:  logFunc,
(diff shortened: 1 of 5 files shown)


mmatczuk:
This is quite inflexible. Also, how do we add separate access log to here?

mmatczuk:
I do not like how the packages (forwarder and httplog) work together - it's not very flexible.
Also, it's not integrated at all with the logging logic we have in middleware - we have diverged.
We need a single flexible solution.

I'm not sure if it's ok to mix http mw with martian modifiers but maybe it's ok since both deal with http anyways.
With that in mind one component can have two applications as http.Handler wrapper and as a modifier.

Going back here I think it would be OK to add [middleware.Logger()](https://github.com/saucelabs/forwarder/blob/61b72918f99dc42dafa99c25c78aa0277519a718/middleware/logger.go#L19) function as a parameter in HTTPServerConfig.
This gives much flexibility and removes the coupling.

httplog can than expose functions that match middleware.Logger signature.
The function selection can be moved to bind module.
It's very easy to write a binding to anything including a function pointer with anyflag library.

Choraden:
I appreciate your suggestions. The code is definitely more flexible and better integrated with middleware components. 

The drawback that's on my mind is that middleware.Logger does not allow to log whole http.Responses (with headers and body). But maybe we don't need to log them? 

My other concern is how to properly count written bytes without delegator. 

PTAL @mmatczuk

httplog/logger.go
```diff
@@ -0,0 +1,181 @@
...
+	ErrOnlyLogMode LoggerMode = "error"
+)
+
+func (m LoggerMode) String() string {
+	return string(m)
+}
+
+// Logger is a modifier that logs requests and responses.
+type Logger struct {
+	log func(line string)
```
mmatczuk, line 36: How about using `Writer io.Writer`?

Then ignore the buffering here! Just use a buffered writer and add separate option for access logs.

mmatczuk: Writer is no go due to to concurrency.

httplog/logger.go
```diff
@@ -0,0 +1,181 @@
...
+// logging the body. Log function defaults to martian.Infof.
+func NewLogger() *Logger {
+	return &Logger{
+		log: func(line string) {
+			log.Infof(line)
+		},
+	}
+}
+
+func (l *Logger) Parse(val string) (middleware.Logger, error) {
```
mmatczuk, line 49: There is a problem with configuration you want to do too much at once. A logger needs a log sink and mode.
This should be read from the config and only then you can properly initialize it...
SetLogFunc should be removed.

Defaulting to martian log is wrong, for instance now HttpServer uses matianlog...

middleware/logger.go
```diff
@@ -28,9 +31,39 @@ func (l Logger) Wrap(h http.Handler) http.Handler {
...
 			Request:  r,
+			Response: nil,
 			Status:   d.Status(),
 			Written:  d.Written(),
 			Duration: time.Since(start),
 		})
 	})
 }
+
+const startTimeKey = "forwarder-middleware-start-time"
```
mmatczuk, line 42: How about `start-time`.

httplog/logger.go
mmatczuk, line 36: How about we used the logging function signature `func(format string, args ...interface{})`

middleware/logger.go
```diff
@@ -28,9 +31,39 @@ func (l Logger) Wrap(h http.Handler) http.Handler {
 		h.ServeHTTP(w, r)
 		l(LogEntry{
 			Request:  r,
+			Response: nil,
```
mmatczuk, line 34: Not needed

Choraden:
I decided to keep LogMode in ServerConfig again and create logger in-place. It was unnatural to initialise logFunc in bind. You would rather take things from config than put them there.

Creating Logger now takes logFunc `func(format string, args ...interface{})` and mode which is more flexible and configurable. 

We could think of creating separate named stdlogger for this, but I don't reckon it's necessary.  

PTAL @mmatczuk

PR #182, opened 2023-01-13
middleware: use prometheus to gather proxy traffic metrics

Choraden:
This patch implements `martian.RequestResponseModifier` interface for Prometheus object, so that it can gather proxy traffic metrics.

### DIFF
--- a/middleware/prometheus.go
+++ b/middleware/prometheus.go
@@ -5,11 +5,13 @@
 package middleware
 
 import (
+	"fmt"
 	"net"
 	"net/http"
 	"strconv"
 	"time"
 
+	"github.com/google/martian/v3"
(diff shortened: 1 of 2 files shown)


middleware/prometheus.go
```diff
@@ -119,3 +121,58 @@ func computeApproximateRequestSize(r *http.Request) int {
 	}
 	return s
 }
+
```
mmatczuk, line 124: I think this is wrong, also the written in delegator does not include the headers size.

Choraden: As we discussed response size metric is skipped for now. 

Later I figured out we could use messageview used in httplog to compute response size if we need it right away. But that would be just a hack (with several drawbacks). It's definitely not the proper way of doing it.

PR #185, opened 2023-01-17
http_proxy: don't use Host header in FindProxyForURL() host parameter

mmatczuk:


e2e/override/proxy-pac-issue-184.yaml
```diff
@@ -0,0 +1,8 @@
+version: "3.8"
+
+services:
+  proxy:
+    environment:
+      FORWARDER_PAC: "/pac.js"
+    volumes:
+      - ./pac/issue-184.js:/pac.js
```
Choraden, line 8: CI does not use it. Is is intended?

http_proxy.go
```diff
@@ -204,7 +204,7 @@ func (hp *HTTPProxy) upstreamProxyURL() *url.URL {
 }
 
 func (hp *HTTPProxy) pacProxy(r *http.Request) (*url.URL, error) {
-	s, err := hp.pac.FindProxyForURL(r.URL, r.Host)
+	s, err := hp.pac.FindProxyForURL(r.URL, "")
```
Choraden, line 207: nit: maybe a short comment on why this is empty

mmatczuk: This is documented in ProxyResolver. I updated the docs and copied them to the interface in forwarder.

e2e/override/proxy-pac-issue-184.yaml
mmatczuk: Good catch, I did not add the run script changes.

Issue #110, opened 2022-10-27
credentials: add support comma separated ports syntax 


mmatczuk:
Like in this example

```
  # Add basic authentication header to requests to example.com and example.org on ports 80 and 443
  forwarder proxy --address localhost:8080 -c bob:bp@example.com:* -c alice:ap@example.org:80,443
```

PR #193, opened 2023-01-20
credentials, config: allow multiple ports syntax parsing

Choraden:
This patch allows comma separated ports in credentials. 
```
Example:
./forwarder --credentials user:pass@host:80,443 - results in two entries of credentials:
1. user:pass@host:80
2. user:pass@host:443
```


Choraden:
I find this solution a little tricky, but I don't see any possible drawbacks, and adding 5 lines of code in parsing logic was too convenient not to do it.

config.go
```diff
@@ -55,9 +55,17 @@ func wildcardPortTo0(val string) string {
 	return strings.Join(s, ":")
 }
 
+var lastUserHost string //nolint:gochecknoglobals // This is for multiple ports parsing.
```
mmatczuk, line 58: This CANNOT be global.

mmatczuk (changes requested):
This is not the best way to approach this problem.

Choraden:
Issue #110 is cancelled.

PR #201, opened 2023-01-26
http proxy: enable connect passthrough

### LINKED ISSUE #196: martian: add Connect passthrough
Please add the following config option to Proxy, and implement.

```
	// ConnectPassthrough handles CONNECT requests as normal requests.
	// The request is passed to RoundTripper and the response is passed back to the client.
	ConnectPassthrough bool
```

Note that the CONNECT request in that form requires flush after write.

This requires a test in martian.

Choraden:
CONNECT requests will be handled as normal request (i.e. round tripped) when passthrough is enabled.

Fixes #196


mmatczuk:
I do not want to expose it like that.

PR #203, opened 2023-01-27
add third party licenses

Choraden:
The licenses are generated with go-licenses tool.
The template is declared in `third_party_license.tpl`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -35,6 +35,12 @@ jobs:
         with:
           go-version: "${{env.GO_VERSION}}"
 
+      - name: Install dependencies
+        run: make install-dependencies
+
+      - name: Verify licenses
+        run: make verify-licenses
+
       - name: Run GoReleaser
         uses: goreleaser/goreleaser-action@v3
         with:
--- a/Makefile
+++ b/Makefile
@@ -10,14 +10,27 @@ include .version
 .PHONY: install-dependencies
 install-dependencies:
 	@rm -Rf bin && mkdir -p $(GOBIN)
-	go install github.com/golangci/golangci-lint/cmd/golangci-lint@$(GOLANGCI_LINT_VERSION)
-	go install github.com/goreleaser/goreleaser@$(GORELEASER_VERSION)
 	go install golang.org/x/tools/cmd/godoc@latest
 	go install golang.org/x/tools/cmd/stringer@latest
+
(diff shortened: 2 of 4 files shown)


Choraden:
@mmatczuk how do you like the format? For me, this is a little too robust. 

Do we want to automate updating these licenses?

Makefile
```diff
@@ -14,6 +14,7 @@ install-dependencies:
 	go install github.com/goreleaser/goreleaser@$(GORELEASER_VERSION)
 	go install golang.org/x/tools/cmd/godoc@latest
 	go install golang.org/x/tools/cmd/stringer@latest
+	go install github.com/google/go-licenses@latest
```
mmatczuk, line 17: Move to GitHub ones.

Actually we could move `golang.org/x/tools` to to top and separate the groups.

mmatczuk:
Sweet. Let's add a makefile target to rerun it.

.github/workflows/go.yml
```diff
@@ -40,3 +40,6 @@ jobs:
 
       - name: Run e2e test
         run: cd e2e && ./run.sh
+
+      - name: verify third party licenses
```
mmatczuk, line 44: This is only relevant in release mode.

mmatczuk:
Could we use just `verify-licenses` for brevity?

.github/workflows/release.yml
```diff
@@ -35,6 +35,9 @@ jobs:
         with:
           go-version: "${{env.GO_VERSION}}"
 
+      - name: Verify licenses
```
mmatczuk, line 38: Please split to 2 steps to eliminate &&

Makefile
```diff
@@ -10,15 +10,28 @@ include .version
...
+	go install github.com/goreleaser/goreleaser@$(GORELEASER_VERSION)
+	go install github.com/google/go-licenses@latest
+
 .PHONY: build
 build:
 	@goreleaser build --rm-dist --snapshot --single-target --skip-validate
 
+.PHONY: gen-licenses
+gen-licenses:
+	./bin/go-licenses report . --template third_party_license.tpl > third_party_licenses.txt
```
mmatczuk, line 26: ./bin/ is not needed and it may be wrong if GOPATH is different.

Choraden: right

mmatczuk (approved):

PR #210, opened 2023-02-02
Expose configuration through api and in debug mode 

### LINKED ISSUE #206: debug log all config options
We want to add a generic system that would allow to see all the configuration options.

A good place to hook it up would be cobra.
For each flag, in the order of appearance in help, print the flag name and the value.
Passwords need to be obfuscated.

Choraden:
This patch exposes flags and theirs values through api and logs them in debug mode. 
All the passwords are redacted.

### DIFF
--- a/bind/flag.go
+++ b/bind/flag.go
@@ -170,3 +170,14 @@ func MarkFlagFilename(cmd *cobra.Command, names ...string) {
 		}
 	}
 }
+
+func DescribeFlags(flags *pflag.FlagSet) string {
+	b := strings.Builder{}
+	flags.VisitAll(func(flag *pflag.Flag) {
+		if flag.Hidden || flag.Name == "help" {
+			return
+		}
+		b.WriteString(fmt.Sprintf("%s: %q\n", flag.Name, flag.Value))
+	})
+	return b.String()
+}
--- a/bind/flag.go
+++ b/bind/flag.go
@@ -5,6 +5,7 @@
 package bind
 
 import (
+	"fmt"
 	"net/url"
 	"os"
 	"strings"
@@ -17,6 +18,20 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 )
+
+func redactURL(u *url.URL) string {
+	return u.Redacted()
+}
+
+func redactUserinfo(ui *url.Userinfo) string {
+	if ui == nil {
+		return ""
+	}
+	if _, has := ui.Password(); has {
+		return fmt.Sprintf("%s:xxxxx", ui.Username())
+	}
+	return ui.Username()
+}
 
 func DNSConfig(fs *pflag.FlagSet, cfg *forwarder.DNSConfig) {
 	fs.VarP(anyflag.NewSliceValue[*url.URL](nil, &cfg.Servers, forwarder.ParseDNSAddress),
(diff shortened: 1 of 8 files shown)


cmd/forwarder/httpbin/httpbin.go
```diff
@@ -23,6 +24,10 @@ func (c *command) RunE(cmd *cobra.Command, args []string) error {
 	}
 	logger := stdlog.New(c.logConfig)
 
+	cmd.Flags().VisitAll(func(f *pflag.Flag) {
+		logger.Infof("%s %s", f.Name, f.Value)
```
mmatczuk, line 28: How about we did type assertions here to host any custom printing logic?

Choraden: It is difficult to break through all the `Value` interfaces and later grab a private field from `anyflag.Value`, but I did it. 
Please have a look.

mmatczuk:
Nice type casting. I think this would be easier:

1. `anyflag.Value` gets `ObfuscatedString() string` function.
1. There is a new constructor that allows to pass an obfuscating function for a type.
1. We configure that in `bind` and here we check if values implement the StringObfuscator ifce
1. Profit

Choraden:
I sent a pr against anyflag https://github.com/mmatczuk/anyflag/pull/1

I decided to solve it in a way which won't require any later type assertions. 
`Redact` Method on `Value` and `SliceValue` allows to define custom String() function.

I chose `Redact` over `Obfuscate` because I wanted it to feel like well known `url.Redacted()`.

mmatczuk:
Obfuscate is even more common, Redact is more precise.

image
image

mmatczuk:
Can make output be more like yaml 

```
forwarder-e2e-upstream-proxy-1  | 2023/02/03 20:25:30 INFO: Configuration
address: ":3128"
cert-file: "" 
```

bind/flag.go
```diff
@@ -32,7 +33,10 @@ func PAC(fs *pflag.FlagSet, pac **url.URL) {
 func HTTPProxyConfig(fs *pflag.FlagSet, cfg *forwarder.HTTPProxyConfig, lcfg *log.Config) {
 	HTTPServerConfig(fs, &cfg.HTTPServerConfig, "", false)
 	LogConfig(fs, lcfg)
-	fs.VarP(anyflag.NewValue[*url.URL](cfg.UpstreamProxy, &cfg.UpstreamProxy, forwarder.ParseProxyURL),
+	redactURL := func(u *url.URL) string {
```
mmatczuk, line 36: Let's define that as a named function so it can be reused.

bind/flag.go
```diff
@@ -101,7 +105,16 @@ func HTTPServerConfig(fs *pflag.FlagSet, cfg *forwarder.HTTPServerConfig, prefix
 		namePrefix+"read-header-timeout", cfg.ReadHeaderTimeout, usagePrefix+"HTTP server read header timeout")
 	fs.DurationVar(&cfg.WriteTimeout,
 		namePrefix+"write-timeout", cfg.WriteTimeout, usagePrefix+"HTTP server write timeout")
-	fs.VarP(anyflag.NewValue[*url.Userinfo](cfg.BasicAuth, &cfg.BasicAuth, forwarder.ParseUserInfo),
+	redactUserinfo := func(ui *url.Userinfo) string {
```
mmatczuk, line 108: Let's define that as a named function so it can be reused.

It would be good to add tests.

bind/flag.go
```diff
@@ -141,3 +154,14 @@ func MarkFlagFilename(cmd *cobra.Command, names ...string) {
 		}
 	}
 }
+
+func DescribeFlags(flags *pflag.FlagSet) string {
+	s := "Configuration\n"
```
mmatczuk, line 159: Not needed.

bind/flag.go
```diff
@@ -141,3 +154,14 @@ func MarkFlagFilename(cmd *cobra.Command, names ...string) {
...
+	flags.VisitAll(func(flag *pflag.Flag) {
+		if flag.Hidden || flag.Name == "help" {
+			return
+		}
+		s += fmt.Sprintf("%s: %q\n", flag.Name, flag.Value)
```
mmatczuk, line 164: How about bytes.Buffer instead.

cmd/forwarder/proxy/proxy.go
```diff
@@ -184,7 +187,16 @@ func Command() (cmd *cobra.Command) {
 	defer func() {
 		fs := cmd.Flags()
 		bind.HTTPProxyConfig(fs, c.httpProxyConfig, c.logConfig)
-		fs.VarP(anyflag.NewSliceValue[*forwarder.HostPortUser](c.credentials, &c.credentials, forwarder.ParseHostPortUser),
+		redactHostPortUser := func(hpu *forwarder.HostPortUser) string {
```
mmatczuk, line 190: This goes to forwarder, as well as parsing.

PR #220, opened 2023-02-09
bind: rename --log-http-requests to --log-http

Choraden:
This patch renames flag to shorter and simpler form.

### DIFF
--- a/e2e/override/proxy-log.yaml
+++ b/e2e/override/proxy-log.yaml
@@ -3,5 +3,5 @@ version: "3.8"
 services:
   proxy:
     environment:
-      FORWARDER_LOG_HTTP_REQUESTS: "true"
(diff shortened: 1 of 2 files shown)


mmatczuk:
Also adjust in `forwarder/e2e/override/proxy-log.yaml`

Choraden:
Good catch!

e2e/override/proxy-log.yaml
```diff
@@ -3,5 +3,5 @@ version: "3.8"
 services:
   proxy:
     environment:
-      FORWARDER_LOG_HTTP_REQUESTS: "true"
+      FORWARDER_LOG_HTTP: "error"
```
mmatczuk, line 6: LOL, this should be headers

PR #257, opened 2023-03-22
Add support for configuration files

### LINKED ISSUE #212: Add support for config files

mmatczuk:
This is a generic implementation with Viper based partially on what we had for env vars.

```
$ go run ./cmd/forwarder -h
HTTP (forward) proxy server with PAC support and PAC testing tools

Commands:
  run           Start HTTP (forward) proxy server
  pac           Tools for working with PAC files

Other Commands:
  completion    Generate the autocompletion script for the specified shell
  version       Print version information

The following options can be passed to any subcommand:

Options:
    --config-file (env FORWARDER_CONFIG_FILE)
        Configuration file to load settings from. The supported formats are: JSON, YAML, TOML, HCL, and Java
        properties. The file format is determined by the file extension, if not specified the default format is YAML.
        The following precedence order of configuration sources is used: command flags, environment variables, config
        file, default values.
```

Fixes #212


cmd/forwarder/root.go
```diff
@@ -19,13 +20,16 @@ import (
 const envPrefix = "FORWARDER"
 
 func rootCommand() *cobra.Command {
+	var configFile string
```
Choraden, line 23: Let's remove this variable, since it is not used anywhere else.

Choraden (commented):
Overall LGTM. 
However, I would exchange priorities  of env vars with config file, as it feels more natural to me. But I understand the order comes from viper, and we don't want to mess with it.

cmd/forwarder/root.go
mmatczuk: Sure.

mmatczuk: bind.ConfigFile(cmd.PersistentFlags(), new(string))

PR #260, opened 2023-03-22
replace verbosity with log levels

Choraden:
This pr replaces verbosity with log levels. Available levels: `error`, `info`, `debug`. Default log level is `info`.


mmatczuk:
Look for verbose in e2e tests.

PR #272, opened 2023-04-07
E2E tests rework 

Choraden:
This pr reworks e2e tests.
- E2E go module has been removed. 
- Bash and Makefile parts have been removed. All the configurations come from pure Golang.
- We no longer use Docker, tests are run on a localhost. Randomising ports allows us to run the test in parallel.  
- A test runner is now responsible for running e2e tests. The test runner takes a test binary and a configuration that needs to be run. It runs the tests in different processes with exec.Command, and saves the outputs of commands in files. 
- Test runner is also capable to dump tests configurations. This makes local test development more straightforward. 
- Linter issues, exposed by removing e2e go module, have been fixed.


mmatczuk:
Squash
- go mod tidy Hubert Grochowski 07/04/2023, 14:35
- e2e: remove go mod Hubert Grochowski 27/03/2023, 08:38

mmatczuk:
Squash 
golangci: fix config Hubert Grochowski 07/04/2023, 17:49
e2e/bench: fix linter issues Hubert Grochowski 07/04/2023, 17:48

mmatczuk:
Consider moving 
ghaction: update e2e workflow Hubert Grochowski 07/04/2023, 14:38 
as the top most commit.

Besides we use `gh:` prefix for everything GitHub related.
It's also worth mention which action you are talking about like in `gh: go, update e2e ...`

.github/workflows/go.yml
```diff
@@ -65,8 +65,15 @@ jobs:
         with:
           go-version: "${{env.GO_VERSION}}"
 
-      - name: Build Docker image
-        run: make update-devel-image
+      - name: Build Forwarder
+        run: go build -o forwarder ./cmd/forwarder
```
mmatczuk, line 69: Use make build instead.

.github/workflows/go.yml
```diff
@@ -65,8 +65,15 @@ jobs:
...
 
       - name: Run e2e test
-        run: cd e2e && ./run.sh
+        run: cd e2e && make test
```
mmatczuk, line 72: make -C e2e

This works as test is the 1st target, you can add `all: test` to make it more explicit in the Makefile.

.github/workflows/go.yml
```diff
@@ -65,8 +65,15 @@ jobs:
...
+
+      - name: Archive test outputs
+        if: always()
```
mmatczuk, line 75: This should only upload if test fails.

mmatczuk:
It would be really nice to have something like [here](https://github.com/marketplace/actions/junit-report-action) to inform users which test failed. Not saying we should use this explicit action but it's one of the possible directions. In this case we need to generate junit.xml then. The benefit is that most CI systems support it.

mmatczuk:
```
	proxy    = flag.String("proxy", "", "URL of the proxy to test against")
	proxyAPI = flag.String("proxy-api", "", "Addr of the proxy API to test against")
	proxyBA  = flag.String("proxy-basic-auth", "", "Basic auth for the proxy (user:password)")

	httpbin    = flag.String("httpbin", "", "URL of the httpbin server to test against")
	httpbinAPI = flag.String("httpbin-api", "", "Addr of the httpbin API to test against")

	upstreamAPI = flag.String("upstream-api", "", "Addr of the upstream API to test against")

	sc2450 = flag.String("sc2450", "", "URL of the SC-2450 server to test against")

	maxWait            = flag.Duration("max-wait", 5*time.Second, "Maximum time to wait for the containers to become ready")
	insecureSkipVerify = flag.Bool("insecure-skip-verify", false, "Skip TLS certificate
```

Could we get all this information from config?

mmatczuk:
Most for the framework should be moved to the `testrunner` pkg shouldn't it?

mmatczuk:
We should ignore e2e/test-outputs/

```
Untracked files:
  (use "git add <file>..." to include in what will be committed)
        e2e/test-outputs/
        forwarder
```

The `forwarder` binary ignore is fixed in main - please rebase.

mmatczuk:
Can we name the output files .log instead of .txt?

mmatczuk:
I suggest we use `Command []string` instead of 

```
	Command    string
	SubCommand string
 ```

**Comment by mmatczuk** [comment]
TODO Add issue to reenable `TestProxyLocalhost`.

**Comment by mmatczuk** [comment]
This is code I generated to avoid using json, you need to convert cases but overall that's it.

```
func listArgs(t interface{}) ([]string, error) {
    v := reflect.ValueOf(t)
    if v.Kind() != reflect.Struct {
        return nil, fmt.Errorf("listArgs only accepts struct types")
    }

    res := make([]string, 0, v.NumField())
    for i := 0; i < v.NumField(); i++ {
        field := v.Type().Field(i)
        fieldName := field.Name
        fieldValue := v.Field(i).Interface()
        if fieldValue == nil {
            continue
        }
        res = append(res, fmt.Sprintf("--%s=%v", fieldName, fieldValue))
    }

    return res, nil
}
```

Given that you can use `Args any` and just feed it to this function.

**Comment by mmatczuk** [comment]
The design of testrunner mixes orthogonal concepts.
We need a library that is documented that can be used to run e2e tests the way we need.
We put all the concrete configurations away from the library into the e2e tests.

**Comment by mmatczuk** [comment]
nit: GetFreePorts(6) is not very idiomatic we should just use GetFreePorts.

File: e2e/testrunner/runner.go [hunk 4]
```diff
@@ -0,0 +1,178 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to list args: %w", err)
+		}
+	}
+	if r.SubCommand != "" {
+		args = append([]string{r.SubCommand}, args...)
+	}
+	cmd := exec.CommandContext(ctx, r.Command, args...) //nolint:gosec // this is a test runner
+	cmd.Stdout = output
+	cmd.Stderr = output
```
mmatczuk, line 51: We should have separate out and err.
Ignore my previous comment on having the .log suffix and let's have .stdout and .stderr for clarity.

mmatczuk:
I think the idea that testrunner/run generates configurations is troublesome.
It would be much cleaner to have a tool to generate test configurations as files.
And another tool to run it.

This way the listing args problem is removed. The passing arguments to test is removed. The test just gets the json file, parses is to a struct and can access any information.
Also args in json should not have `--` prefix.

.github/workflows/go.yml
Choraden: I didn't use makefile, because it was broken and I forgot about it. Here's the fix https://github.com/saucelabs/forwarder/pull/277. I should've done it in the first place.

e2e/testrunner/config.go
```diff
@@ -0,0 +1,91 @@
...
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"gopkg.in/yaml.v3"
+)
+
+type Config struct {
+	Runnable
+	ConfigFile any
```
mmatczuk, line 19: This is quite confusing, ConfigFile should be a file not any.
I think we could add `Config any` to `Runnable`, this would further streamline it.

e2e/testrunner/runner.go
```diff
@@ -0,0 +1,154 @@
...
+	Command []string `yaml:"command,omitempty"`
+}
+
+func (r *Runnable) RunAndSaveOutput(ctx context.Context, path string) error {
+	if r.Command == nil || len(r.Command) == 0 {
+		return fmt.Errorf("no command to run for %s", r.Name)
+	}
+
+	cmd := exec.CommandContext(ctx, r.Command[0], r.Command[1:]...) //nolint:gosec // this is a test runner
+	stdout, stderr := &bytes.Buffer{}, &bytes.Buffer{}
```
mmatczuk, line 36: Why not var?

e2e/testrunner/runner.go
```diff
@@ -0,0 +1,154 @@
...
+		return
+	}
+}
+
+type Configuration struct {
+	Name      string
+	Runnables []Runnable
+}
+
+type RunnerConfig struct {
```
mmatczuk, line 62: Inline this type and just pass params to constructor - it's just 2 fields.

e2e/testrunner/runner.go
```diff
@@ -0,0 +1,154 @@
...
+
+func (r *Runner) run() error {
+	return filepath.WalkDir(r.Root, func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if r.Root == path || !d.IsDir() {
+			return nil
+		}
+		b, err := os.ReadFile(filepath.Join(path, "run.yaml"))
```
mmatczuk, line 91:
```
- name: test
  command:
    - ./e2e.test
    - --config-file=test-data/default-http-http-http/test.yaml
```
It oddly does not fit here 🤔

Choraden:
Closing in favour of the new approach #278.

PR #277, opened 2023-04-17
makefile: export shell with overwritten path

Choraden:
By default Makefile executes lines in a separate shell invocations. Changing PATH variable in one line does not affect next invocations.

Before this patch Makefile would not use executables from '/bin' directory as assigning to PATH wasn't persistent. 
```bash
make build
make: goreleaser: No such file or directory
make: *** [build] Error 1
```

This patch fixes it, by exporting SHELL that uses up-to-date PATH.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@
 
 export GOBIN ?= $(PWD)/bin
 export PATH  := $(GOBIN):$(PATH)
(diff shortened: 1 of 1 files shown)


Choraden:
You are right. Looks like a bug in a make 😕. Update fixed it.

PR #278, opened 2023-04-21
Rework e2e tests 

Choraden:
The second version of E2E tests rework. 
This approach is based on running test in docker, one at a time. This eliminates a lot of issues, e.g. randomising ports, complex logic to run tests in parallel, localhost tests. 

Each test consists of docker compose and a callback. The callback function is executed, once the compose is up and running. 

When the test fails user will find the current setup at `/e2e/docker-compose.yaml`, containers aren't purged after test, so one can use them as well. Logs can be retrieved with `make dump-logs`. 

E2E test now uses environment variables to retrieve information about current configuration.


e2e/testrunner/runner.go
```diff
@@ -0,0 +1,68 @@
...
+	if err := cmd.Run(); err != nil {
+		log.Printf("compose stdout: %s", stdout.String())
+		log.Printf("compose stderr: %s", stderr.String())
+		return fmt.Errorf("compose up, test %s: %w", t.Name, err)
+	}
+	return nil
+}
+
+func (t *Test) Run() error {
+	log.Printf("Running test %s\n", t.Name)
```
mmatczuk, line 41:
```suggestion
	log.Printf("Starting environment %s\n", t.Name)
```

e2e/testrunner/runner.go
```diff
@@ -0,0 +1,68 @@
...
+	if err := t.composeUp(); err != nil {
+		return err
+	}
+	if t.Callback != nil {
+		if err := t.Callback(); err != nil {
+			return err
+		}
+	} else {
+		return fmt.Errorf("test %s: no callback", t.Name)
```
mmatczuk, line 50: This is not test

e2e/README.md
```diff
@@ -1,22 +1,23 @@
 # Forwarder e2e tests
```
mmatczuk, line 1: Please add information on debugging and use of make when something fails.

.github/workflows/go.yml
```diff
@@ -69,4 +69,8 @@ jobs:
         run: make update-devel-image
 
       - name: Run e2e test
-        run: cd e2e && ./run.sh
+        run: cd e2e && make run-e2e
```
mmatczuk, line 72: Use make -C e2e instead.

.github/workflows/go.yml
```diff
@@ -69,4 +69,8 @@ jobs:
...
+
+      - name: Dump Logs
+        if: failure()
+        run: cd e2e && make dump-logs
```
mmatczuk, line 76: Same here.

.golangci.yml
```diff
@@ -83,7 +83,7 @@ issues:
       linters:
         - gochecknoglobals
 
-    - path: _test\.go
+    - path: _test.go
```
mmatczuk, line 86: ?

e2e/Makefile
```diff
@@ -33,7 +31,19 @@ test: e2e.test
 	@docker run --name "test-runner" --network "forwarder-e2e_default" --cpuset-cpus 0 \
 	-v "$(PWD)/e2e.test:/usr/bin/e2e.test" -i --read-only --rm \
 	$(shell docker inspect forwarder-e2e-proxy-1 --format='{{ range .Config.Env }}{{ if eq "FORWARDER" (slice . 0 9) }} -e {{.}}{{end}}{{end}}') \
-	ubuntu e2e.test -test.run $(RUN) $(ARGS)
+	-e HTTPBIN=$(HTTPBIN) -e INSECURE=$(INSECURE) -e TEST_SC=$(TEST_SC) ubuntu e2e.test -test.run $(RUN) -test.v $(ARGS)
```
mmatczuk, line 34: TEST_SC again?

mmatczuk: The sc thing is irrelevant it should be part of the compose.

e2e/framework.go
```diff
@@ -16,21 +22,20 @@ import (
 	"github.com/gorilla/websocket"
 )
 
+func h2ToHTTP(p string) string {
```
mmatczuk, line 25: This should be like given env var name return scheme.

e2e/testrunner/compose.go
```diff
@@ -0,0 +1,89 @@
...
+	Image       string            `yaml:"image,omitempty"`
+	Command     string            `yaml:"command,omitempty"`
+	Environment map[string]string `yaml:"environment,omitempty"`
+	Ports       []string          `yaml:"ports,omitempty"`
+	Volumes     []string          `yaml:"volumes,omitempty"`
+}
+
+const indent = "  "
+
+func (svc *Service) ToStringSlice() ([]string, error) {
```
mmatczuk, line 30: I suggest inlining this function.

e2e/testrunner/compose.go
```diff
@@ -0,0 +1,89 @@
...
+	return s, nil
+}
+
+type Compose []*Service
+
+func NewCompose(services ...*Service) Compose {
+	return services
+}
+
+const version = "\"3.8\""
```
mmatczuk, line 49: composeVersion

e2e/testrunner/compose.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	s := strings.Split(string(b), "\n")
+	for i := range s {
+		s[i] = indent + s[i]
+	}
+	s = append([]string{svc.Name + ":"}, s...)
+	return s, nil
+}
+
+type Compose []*Service
```
mmatczuk, line 43: Let's add a proper struct. It can be extended unlike this.

e2e/testrunner/compose.go
mmatczuk: Also version can be Compose field.

e2e/testrunner/compose.go
```diff
@@ -0,0 +1,89 @@
...
+
+func NewCompose(services ...*Service) Compose {
+	return services
+}
+
+const version = "\"3.8\""
+
+func (c Compose) Marshal() ([]byte, error) {
+	b := &bytes.Buffer{}
```
mmatczuk, line 52: use var

e2e/tests.go
```diff
@@ -0,0 +1,298 @@
...
+	"github.com/saucelabs/forwarder/e2e/testrunner"
+)
+
+const (
+	proxyService    = "proxy"
+	upstreamService = "upstream-proxy"
+	HttpbinService  = "httpbin"
+)
+
+type testMod func(t *testrunner.Test)
```
mmatczuk, line 26: Can we use TestOpt instead?

e2e/tests.go
```diff
@@ -0,0 +1,298 @@
...
+			}
+			return err
+		}
+	}
+}
+
+func withProtocols(proxy, httpbin string) testMod {
+	return func(t *testrunner.Test) {
+		for _, s := range t.Compose {
+			if s.Name == "proxy" {
```
mmatczuk, line 65: Use const

e2e/tests.go
```diff
@@ -0,0 +1,298 @@
...
+			} else if s.Name == "httpbin" {
+				s.Environment["FORWARDER_PROTOCOL"] = httpbin
+			}
+		}
+	}
+}
+
+func withBasicAuth(auth string) testMod {
+	return func(t *testrunner.Test) {
+		for _, s := range t.Compose {
```
mmatczuk, line 76: Looking at how you have to iterate Compose should have a Service(name string) helper.

e2e/testrunner/compose.go
```diff
@@ -0,0 +1,89 @@
...
+	for i := range s {
+		s[i] = indent + s[i]
+	}
+	s = append([]string{svc.Name + ":"}, s...)
+	return s, nil
+}
+
+type Compose []*Service
+
+func NewCompose(services ...*Service) Compose {
```
mmatczuk, line 45: Not used.

e2e/testrunner/runner.go
```diff
@@ -0,0 +1,68 @@
...
+package testrunner
+
+import (
+	"bytes"
+	"fmt"
+	"log"
+	"os/exec"
+)
+
+type Test struct {
```
mmatczuk, line 16: When you use that testrunner.Test is confusing I suggest migrating all of it to Compose then it would be much simpler.

e2e/testrunner/runner.go
```diff
@@ -0,0 +1,68 @@
...
+	Name string
+	// Callback function is called after compose is up and running.
+	Callback func() error
```
mmatczuk, line 19: Maybe rename it to OnStart so that the comment is not needed / evident.

mmatczuk:
My general comments

* testrunner should expose only Compose and it should have all the lifecycle functions.
* testrunner.Compose must have teardown function i.e. how do we make sure we clear the last environment?
* I do not like the defaultTest structure, there is too many shortcuts, the purpose is to make it flexible. Now it's not. We don't need upstream all the time and how do I add test with 7 upstreams? Not I like the name of it.
* *Compose lacks helper functions like AddService(name, image, opts...) that would help build it with an elegant DSL* 

My suggestions

* testrunner.Compose chould have Path with docker-compose file path - by default it can be temp file.
* testrunner.Service could have a Wait(s *service) function that would do the waiting then it would be easier to integrate in the test output.

e2e/tests.go
```diff
@@ -0,0 +1,298 @@
...
+	"os/exec"
+	"strings"
+
+	"github.com/saucelabs/forwarder/e2e/testrunner"
+)
+
+const (
+	proxyService    = "proxy"
+	upstreamService = "upstream-proxy"
+	HttpbinService  = "httpbin"
```
mmatczuk, line 23: Why exported?

mmatczuk:
Use of the word test(s) is very confusing like in tests.go - those are not tests but test environments / setups / you name it.

Choraden:
@mmatczuk I've addressed your comments. PTAL.

.gitignore
```diff
@@ -51,3 +51,5 @@ tags
 
 # scratch repo to hack
 /scratch
+
+/e2e/docker-compose.yaml
```
mmatczuk, line 55: This should be done in the e2e directory.

mmatczuk:
Can we move `setups.go` to `run` pkg?

e2e/run/run.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"log"
+
+	"github.com/saucelabs/forwarder/e2e"
+)
+
+func main() {
+	for _, setup := range e2e.AllSetups() {
+		if err := setup.Run(); err != nil {
+			log.Printf("FAIL: %v", err)
```
mmatczuk, line 18: log.Fatal and make sure we exit with appropriate status code.

mmatczuk:
I'm tempted to rename testrunner to compose

Choraden:
> Can we move `setups.go` to `run` pkg?

done

Choraden:
> I'm tempted to rename testrunner to compose

so am I

mmatczuk (approved):
LGTM!

PR #279, opened 2023-05-04
Remove httpexpect module from E2E tests 

Choraden:
This patch exchanges httpexpect module for manually implemented substitute with minimal scope.
New Client is light, handy, and easily extendable.


e2e/e2e_test.go
```diff
@@ -37,34 +39,42 @@ func TestStatusCodes(t *testing.T) {
 		500, 501, 502, 503, 504, 505, 506, 507, 508, 510, 511, 599,
 	}
 
-	e := Expect(t, httpbin)
```
mmatczuk, line 40: I suggest not changing the name to minimize the patch size.

mmatczuk: n/m this is better I guess.

PR #280, opened 2023-05-05
Headers modifiers tests and improvements 

Choraden:
- Env variables copied from containers are now escaped as there were issues with `;` char
- `Header::String()` method will now guard printing headers in the right format
- Header flag values are now escaped as there were issues with `*` char
- `e2e.Client` and `e2e.Request` logging is improved
- Tests of request/response headers modifiers 
  - To test the headers HTTPBin was extended with a headerHandler.
    Header handler implements the `/header/{key}/{value}` endpoint.
    It validates whether request contains the header with the given key and value.
    It will also send the given header back, so one can test response modifier as well.


header/header.go
```diff
@@ -95,6 +95,21 @@ func removeHeadersByPrefix(h http.Header, prefix string) {
...
+	case Remove:
+		return "-" + h.Name
+	case RemoveByPrefix:
+		return "-" + h.Name + "*"
+	case Empty:
+		return h.Name + ";"
+	case Add:
+		return h.Name + ":" + *h.Value
+	default:
+		return "unknown action"
```
mmatczuk, line 109: why not ""

mmatczuk:
This is nice.

mmatczuk (approved):

header/header.go
Choraden: Fixed.

PR #291, opened 2023-05-23
Move NewProxyURL from e2e to httpexpect and some improvements 

Choraden:


utils/httpexpect/httpexpect.go
```diff
@@ -8,12 +8,44 @@ package httpexpect
...
+	if basicAuth != "" {
+		user, pass, _ := strings.Cut(basicAuth, ":")
+		u.User = url.UserPassword(user, pass)
+		tb.Log("using basic auth for proxy", u)
+	}
+
+	return http.ProxyURL(u)
+}
+
+func NewTransport(tb testing.TB, insecure bool, proxy func(*http.Request) (*url.URL, error)) *http.Transport {
```
mmatczuk, line 37: Here tb is not used

utils/httpexpect/httpexpect.go
```diff
@@ -8,12 +8,44 @@ package httpexpect
...
+	"crypto/tls"
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
+	"strings"
 	"testing"
 )
 
+func NewProxy(tb testing.TB, proxyURL, basicAuth string) func(*http.Request) (*url.URL, error) {
```
mmatczuk, line 20: Rename to ProxyURL, remove tb param.

utils/httpexpect/httpexpect.go
mmatczuk: I suggest not to include it in the lib.

mmatczuk:
LGTM.
WDYT about renaming framework.go to client.go?

PR #292, opened 2023-05-25
logs improvements

mmatczuk:


mmatczuk:
cc @Choraden

Choraden:
@mmatczuk shorthand "k" should be also removed from README.

PR #294, opened 2023-05-30
e2e: improve developer experience and readability

mmatczuk:


Choraden:
I like it so much. Good job.

mmatczuk:
Adding exponential backoff is probably not worth it.
To be clear you want get the info about being ready to test asap.
Adding long polling to readiness would be more aligned with this goal.

mmatczuk:
We can open a new issue to address this issue.

PR #309, opened 2023-07-03
e2e wait refactoring

mmatczuk:


e2e/run/setups/wait.go
```diff
@@ -29,25 +36,19 @@ func WaitForServerReady(baseURL string) error {
...
-	for i := 0; i < int(maxWait/backoff); i++ {
-		resp, rerr = client.Do(req.Clone(context.Background()))
+	deadline := time.Now().Add(MaxWait)
+	for {
+		resp, err := client.Do(req.Clone(context.Background()))
 
 		if resp != nil && resp.StatusCode == http.StatusOK {
 			resp.Body.Close() //noline:errcheck // we don't care about the body
 			return nil
 		}
```
Choraden, line 46: BTW Aren't we leaking response body in here? We should close it regardless the status code. Could you double check?

e2e/wait/wait.go
```diff
@@ -0,0 +1,71 @@
...
+	Client http.Client
+	// Endpoint is the endpoint to check for readiness.
+	Endpoint string
+	// MaxWait is the maximum time to wait for a server to be ready.
+	MaxWait time.Duration
+	// Backoff is the time to wait between checks.
+	Backoff time.Duration
+}
+
+var DefaultWaiter = Waiter{
```
Choraden, line 27: Do we need the `DefaultWaiter` to be exported now? How about we make it a function. 

We will use it to make a local variable `defaultWaiter` in the package for `ServerReady` function. 
Users who would like to change the waiter config could call this function and modify the struct.

e2e/wait/wait.go
```diff
@@ -0,0 +1,71 @@
...
+			if resp.StatusCode == http.StatusOK {
+				return nil
+			}
+		}
+
+		if time.Now().After(deadline) {
+			if err != nil {
+				return fmt.Errorf("not ready after %s: %w", w.MaxWait, err)
+			}
+			if resp.StatusCode != http.StatusOK {
```
Choraden, line 59: We already know this condition is true. Is it worth checking it?

e2e/wait/wait.go
mmatczuk: I guess it does not matter. This way it's possible to change the defaults.
I can unexpert it no problem.

PR #310, opened 2023-07-04
e2e setup refactoring

mmatczuk:


mmatczuk:
I turns out it serves no purpose.

PR #316, opened 2023-07-17
fix setting dial function in proxy

Choraden:
Martian has an interwinted logic for setting http.Transport and the dialer. 
Call to SetRoundTripper overwrites tr.DialContext with hp.proxy.dial. 
When hp.transport is a pointer to http.Transport the assignment is visible in the forwarder code. 
The dialer needs to be set first to avoid overwriting.


mmatczuk:
I'd suggest to fix that in Martian, the API is impossible to get right.

- SetDialContext should be removed.
- SetRoundTripper should wrap the dial function like SetDialContext for http.Transport
- The behavior must be documented

mmatczuk (approved):

PR #320, opened 2023-07-28
dns: fix custom servers support and extend tests 

Choraden:


Choraden:
It looks like the dns fallback mechanism is broken. 
```go
func (r *resolver) dialDNS(ctx context.Context, network, address string) (net.Conn, error) {
	for _, u := range r.servers {
		r.log.Debugf("dial DNS server %s instead of %s", u, address)
		conn, err := r.dialer.DialContext(ctx, network, u.String())
		if err != nil {
			r.log.Errorf("failed to dial DNS server %s: %v", u, err)
			continue
		}
		return conn, nil
	}

	return nil, fmt.Errorf("failed to dial DNS server")
}
```
Custom dialDNS function always returns the connection to the first dns server, even if the server is wrong on purpose. It looks like dial to UDP succeeds every time, but I have no idea why.

The error is encountered later when dns query is performed. 
```
2023/07/28 13:16:34 INFO: using DNS servers: [127.0.0.1:34567 1.1.1.1:53]
--- FAIL: TestResolverLookupHostFallback (0.00s)
    dns_test.go:73: LookupHost failed: lookup google.com on 10.255.3.12:53: read udp 127.0.0.1:61671->127.0.0.1:34567: read: connection refused
    dns_test.go:76: LookupHost returned no address
```

bind/flag.go
```diff
@@ -39,7 +39,8 @@ func DNSConfig(fs *pflag.FlagSet, cfg *forwarder.DNSConfig) {
 			"If specified multiple times, the first one is used as primary server, the rest are used as fallbacks. "+
 			"The port is optional, if not specified the default port is 53. ")
 	fs.DurationVar(&cfg.Timeout,
-		"dns-timeout", cfg.Timeout, "Timeout for dialing DNS servers.")
+		"dns-timeout", cfg.Timeout, "Timeout for dialing DNS servers. "+
+			"If specified multiple servers, the timeout for each lookup is equal to timeout divided by the number of servers.")
```
mmatczuk, line 43: This should be for each server.

dial.go
```diff
@@ -0,0 +1,148 @@
...
+	for _, s := range cfg.DNSServers {
+		dnsAddress := s.String()
+		resolvers = append(resolvers, net.Resolver{
+			PreferGo: true,
+			Dial: func(ctx context.Context, network, address string) (net.Conn, error) {
+				l.Debugf("dial DNS server %s instead of %s", dnsAddress, address)
+				return nd.DialContext(ctx, network, dnsAddress)
+			},
+		})
+	}
```
mmatczuk, line 96: Refactor to function, if no DNS servers specified return nil

dial.go
```diff
@@ -0,0 +1,148 @@
...
+func (d *Dialer) LookupIP(ctx context.Context, network, hostname string) ([]net.IP, error) {
+	for i := range d.resolvers {
+		lctx, cancel := context.WithTimeout(ctx, d.cfg.ResolveTimeout)
+		ips, err := d.resolvers[i].LookupIP(lctx, network, hostname)
+		cancel()
+		if err != nil {
+			d.log.Debugf("failed to resolve %s on %s: %s", hostname, d.cfg.DNSServers[i], err)
+		} else {
+			return ips, nil
+		}
```
mmatczuk, line 115: Refactor to function

dial.go
```diff
@@ -0,0 +1,148 @@
...
+		if err != nil {
+			d.log.Debugf("failed to dial %s: %s", address, err)
+		} else {
+			return conn, nil
+		}
+	}
+	return nil, fmt.Errorf("failed to dial %s on any IP address", address)
+}
+
+func (d *Dialer) DialFunc() dialFunc { //nolint:revive,golint // Unexported by design.
```
mmatczuk, line 143: I agree with the linter. It looks odd and has no purpose really. The Dialer should implement DialContext.

dial_test.go
```diff
@@ -0,0 +1,78 @@
...
+	"net/netip"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/saucelabs/forwarder/log/stdlog"
+)
+
+func TestDialerLookupHost(t *testing.T) {
+	if _, ok := os.LookupEnv("CI"); ok {
```
mmatczuk, line 20: We should run in in CI. We have tests talking to google.com so a DNS query to CF won't hurt.

dial.go
```diff
@@ -0,0 +1,148 @@
...
+		KeepAlive: cfg.KeepAlive,
+	}
+
+	resolvers := make([]net.Resolver, 0, len(cfg.DNSServers))
+	for _, s := range cfg.DNSServers {
+		dnsAddress := s.String()
+		resolvers = append(resolvers, net.Resolver{
+			PreferGo: true,
+			Dial: func(ctx context.Context, network, address string) (net.Conn, error) {
+				l.Debugf("dial DNS server %s instead of %s", dnsAddress, address)
```
mmatczuk, line 92: What if I have 5 dns servers?
Would it fallback each of them?

dial_test.go
```diff
@@ -16,7 +16,7 @@ import (
 	"github.com/saucelabs/forwarder/log/stdlog"
 )
 
-func TestResolverLookupHost(t *testing.T) {
```
mmatczuk: I'd like to see a test for context cancelation.

mmatczuk:
I'm still not really convinced to using the resolver hack is better.

For instance https://github.com/rs/dnstrace provides you with more robust client supporting recursive queries (with CNAME support).
On top of that we would get DNS over https support for free.

The full list is here maybe you can find something even better https://pkg.go.dev/github.com/miekg/dns?tab=importedby.

dial.go
```diff
@@ -0,0 +1,148 @@
...
+	return nil, fmt.Errorf("failed to resolve %s on any DNS server", hostname)
+}
+
+func (d *Dialer) DialWithLookup(ctx context.Context, network, address string) (net.Conn, error) {
+	host, port, err := net.SplitHostPort(address)
+	if err != nil {
+		return nil, err
+	}
+
+	ips, err := d.LookupIP(ctx, "ip", host)
```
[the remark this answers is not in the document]
Choraden: http RoundTripper uses `tcp` network, when resolver.LookupIP needs `ip`, `ipv4` or `ipv6`

dial.go
Choraden: It would sequentially dial the 5 servers, one after another, unless a lookup succeeds. The net.Dialer dials two times by default. 
```
[proxy] DEBUG: dial DNS server 192.168.100.13:53 instead of 127.0.0.11:53
[proxy] DEBUG: dial DNS server 192.168.100.13:53 instead of 127.0.0.11:53
[proxy] DEBUG: failed to resolve wronghost on 192.168.100.13:53: lookup wronghost on 127.0.0.11:53: no such host
[proxy] ERROR: failed to round trip: lookup : failed to resolve wronghost on any DNS server
```

dial_test.go
Choraden: Good idea.

Choraden:
I agree, the resolver hack has its disadvantages, but I would not give up this idea. We can try with the client and compare the results.

Choraden:
We abandoned the custom dialer with resolvers idea in favour of using Golang's net package internals. The internals are hidden from user. To overcome it, we link the private function and manually configure dnsConfig. This way we can manipulate the behaviour of internal Golang's Resolver.

dns.go
```diff
@@ -7,33 +7,87 @@
...
 	"net/netip"
+	"sync"
+	"sync/atomic"
 	"time"
+	_ "unsafe" // for go:linkname
 
 	"github.com/saucelabs/forwarder/log"
 )
 
+// dnsConfig is 1:1 copy of net.dnsConfig struct.
```
mmatczuk, line 20: put it to a separate file then.

bind/flag.go
```diff
@@ -104,12 +94,39 @@ func Credentials(fs *pflag.FlagSet, credentials *[]*forwarder.HostPortUser) {
...
+		namePrefix+"dial-timeout", cfg.DialTimeout,
 		"The maximum amount of time a dial will wait for a connect to complete. "+
 			"With or without a timeout, the operating system may impose its own earlier timeout. For instance, TCP timeouts are often around 3 minutes. ")
 
+	DNSConfig(fs, &cfg.DNSConfig)
+}
+
+func HTTPTransportConfig(fs *pflag.FlagSet, cfg *forwarder.HTTPTransportConfig) {
+	DialConfig(fs, &cfg.DialConfig, "http")
+
```
mmatczuk, line 129: Function order.

dial.go
```diff
@@ -0,0 +1,63 @@
...
+	// probes for an active network connection.
+	// If zero, keep-alive probes are sent with a default value
+	// (currently 15 seconds), if supported by the protocol and operating
+	// system. Network protocols or operating systems that do
+	// not support keep-alives ignore this field.
+	// If negative, keep-alive probes are disabled.
+	KeepAlive time.Duration
+}
+
+func DefaultDialConfig() DialConfig {
```
mmatczuk, line 34: Could we return a ptr here?

dial.go
```diff
@@ -0,0 +1,63 @@
...
+		KeepAlive:   30 * time.Second,
+	}
+}
+
+type Dialer struct {
+	cfg DialConfig
+	nd  *net.Dialer
+}
+
+func NewDialer(cfg DialConfig) (*Dialer, error) {
```
mmatczuk, line 46: Same here pass a ptr, Dialer keeps a copy like it does now.

utils/osdns/dnsconf.go
```diff
@@ -0,0 +1,51 @@
+// Copyright 2023 Sauce Labs Inc. All rights reserved.
```
mmatczuk, line 1: It looks like this file should also have the Go license in it.

It would be also good to point to the file it's based on.

utils/osdns/resolvconf.go
```diff
@@ -0,0 +1,41 @@
+// Copyright 2023 Sauce Labs Inc. All rights reserved.
```
mmatczuk, line 1: Same as in dnsconf.go

bind/flag.go
```diff
@@ -32,14 +33,20 @@ func ConfigFile(fs *pflag.FlagSet, configFile *string) {
 			"The following precedence order of configuration sources is used: command flags, environment variables, config file, default values. ")
 }
 
-func DNSConfig(fs *pflag.FlagSet, cfg *forwarder.DNSConfig) {
+func DNSConfig(fs *pflag.FlagSet, cfg *osdns.Config) {
 	fs.VarP(anyflag.NewSliceValue[netip.AddrPort](nil, &cfg.Servers, forwarder.ParseDNSAddress),
 		"dns-server", "n", "<ip>[:<port>]"+
 			"DNS server(s) to use instead of system default. "+
-			"If specified multiple times, the first one is used as primary server, the rest are used as fallbacks. "+
+			"If specified multiple times, the first one is used as primary server, the rest are used as fallbacks, "+
```
mmatczuk, line 40: Would be good to add a note on --dns-round-robin

bind/flag.go
```diff
@@ -32,14 +33,20 @@ func ConfigFile(fs *pflag.FlagSet, configFile *string) {
...
+			"unless the round-robin option is enabled. "+
 			"The port is optional, if not specified the default port is 53. ")
+
 	fs.DurationVar(&cfg.Timeout,
-		"dns-timeout", cfg.Timeout, "Timeout for dialing DNS servers.")
+		"dns-timeout", cfg.Timeout, "Timeout for dialing DNS servers. "+
+			"Only used if DNS servers are specified. ")
+
+	fs.BoolVar(&cfg.RoundRobin, "dns-round-robin", cfg.RoundRobin, "Enable round-robin DNS server selection. "+
+		"Won't have any effect if DNS servers are not specified. ")
```
mmatczuk, line 49: This sentence is adds no value. Add something more relevant

If more than one DNS server is specified with the --dns-server flag, passing this flag ....

cmd/forwarder/pac/server/server.go
```diff
@@ -36,7 +38,14 @@ func (c *command) RunE(cmd *cobra.Command, args []string) error {
 	logger := stdlog.New(c.logConfig)
 	logger.Debugf("configuration\n%s", config)
 
-	t, err := forwarder.NewHTTPTransport(c.httpTransportConfig, nil)
+	if len(c.dnsConfig.Servers) > 0 {
+		logger.Named("dns").Infof("using DNS servers %v", c.dnsConfig.Servers)
```
mmatczuk, line 42: It would be nicer using strings.Join(c.dnsConfig.Servers, ", ",)

e2e/setups.go
```diff
@@ -202,41 +202,56 @@ func SetupFlagResponseHeader(l *setupList) {
...
 func SetupFlagDNSServer(l *setupList) {
 	const (
-		networkName   = "forwarder-e2e_default"
-		httpbinIPAddr = "192.168.100.10"
-		proxyIPAddr   = "192.168.100.11"
-		dnsIPAddr     = "192.168.100.13"
+		networkName      = "forwarder-e2e_default"
+		httpbinIPAddr    = "192.168.100.10"
+		proxyIPAddr      = "192.168.100.11"
+		dnsIPAddr        = "192.168.100.13"
```
mmatczuk, line 208: It would be nicer if DNS came first with lower IPs.

mmatczuk (approved):
Those can be squashed.

e2e(dns server setup): increase http dial timeout for proxy Hubert Grochowski 04/08/2023, 09:37
e2e/forwarder(service): add withHTTPDialTimeout method Hubert Grochowski 04/08/2023, 09:35


Overall very well done, work nicely split to commits.

I left some minor comments inline.

LGTM

Choraden:
@mmatczuk I've addressed your comments. Merge when you are ready.

Choraden:
@mmatczuk I also tweaked sc-2450 setup health check values as the server got frozen in the previous run.

PR #324, opened 2023-08-01
utils/cobrautil/templates: new value type search algorithm

mmatczuk:
This patch replaces regexp with manual searching for matching bracket. The regexp implementation is not aware of the brackets semantics and matches the last closing bracket preventing the use of <> and [] later in usage. The current implementation finds the value type definition until the first upper case letter.

### DIFF
--- a/utils/cobrautil/templates/help_flags_printer.go
+++ b/utils/cobrautil/templates/help_flags_printer.go
@@ -20,8 +20,8 @@ import (
 	"bytes"
 	"fmt"
 	"io"
-	"regexp"
 	"strings"
+	"unicode"
 
 	"github.com/mitchellh/go-wordwrap"
 	flag "github.com/spf13/pflag"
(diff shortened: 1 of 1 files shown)


Choraden:
I was just wondering what messed up the pac flag description

Choraden:
Basing on the assumption of uppercase letter might be error prone. Maybe we could some document this behaviour in `bind` package.

utils/cobrautil/templates/help_flags_printer.go
```diff
@@ -95,13 +95,12 @@ func writeFlag(out io.Writer, f *flag.Flag, envPrefix string) {
 	fmt.Fprintf(out, getFlagFormat(f), f.Shorthand, f.Name, name, def, env, usage, deprecated)
 }
 
-var valueTypeRegex = regexp.MustCompile(`^[<|\[].+[>|\]]`)
-
 func flagNameAndUsage(f *flag.Flag) (string, string) {
```
Choraden, line 98: Why is it called NameAndUsage? This should be ValueAndUsage, shouldn't it?

mmatczuk: Library naming.

mmatczuk:
The alternative is invent a magic char sequence and apply it as a marker to every usage.
For example "\n" could be good choice but you have to keep it in mind or maintain it which is even more error prone.

The uppercase is fine under the assumption that you always begin sentence with an upper case letter.

Waiting for more concrete suggestions.

Choraden:
I had a very similar idea with a magic char `'`. Moreover, we could force users to implicitly define value type. This is return error when value is not matched.

mmatczuk:
I prefer to limit scope of the changes to the docs.

PR #329, opened 2023-08-03
e2e: tests for insecure

mmatczuk:


e2e/forwarder/service.go
```diff
@@ -81,6 +81,17 @@ func (s *Service) WithProtocol(protocol string) *Service {
 	return s
 }
 
+func (s *Service) SelfSigned() *Service {
+	delete(s.Environment, "FORWARDER_CERT_FILE")
+	delete(s.Environment, "FORWARDER_KEY_FILE")
```
Choraden, line 86: I'm not a fan of this approach. It forces user to keep the right order of options, otherwise it can break. However, when used with caution, maybe it's not that bad.

e2e/setups.go
```diff
@@ -107,6 +108,42 @@ func AuthSetups() (ss []setup.Setup) {
 	return
 }
 
+func UpstreamSetups() []setup.Setup {
```
Choraden, line 111: I got a little confused with the setups naming. Part of the DefaultsSetups is also an Upstream. 

Maybe this is an `--insecure` flag test?

http_error.go
```diff
@@ -26,45 +27,94 @@ const ErrorHeader = "X-Forwarder-Error"
...
+	if ok := errors.As(err, &certErr); ok {
+		code = http.StatusBadGateway
+		msg = "TLS handshake failed"
+	}
+
+	return
+}
+
+func handleDenyError(req *http.Request, err error) (code int, msg string) {
+	if _, ok := cause(err).(denyError); ok { //nolint:errorlint // makes no sense here
```
Choraden, line 96: we could use `errors.As` here as well

Choraden: It doesn't matter, since we need to keep `cause(err)` function for net.Error

mmatczuk: All code uses errors pkg now.

Choraden:
Overall looking good.

e2e/forwarder/service.go
mmatczuk:
```
				AddService(
					forwarder.UpstreamProxyService().
						WithSelfSigned("https")).
```

e2e/setups.go
mmatczuk: I'd change that to a flag test.

mmatczuk:
Comments are addressed.
I added a commit on top to manage the setups better.

@Choraden PTAL

e2e/forwarder/service.go
Choraden: It's definitely better. It would be perfect if there was an assertion, but it is not necessary.

http_error.go
Choraden: great

Choraden (approved):
LGTM

PR #335, opened 2023-08-08
MITM

### LINKED ISSUE #145: proxy enable mitm
TBD

mmatczuk:
Fixes #145

### DIFF
--- a/bind/flag.go
+++ b/bind/flag.go
@@ -151,8 +151,8 @@ func TLSClientConfig(fs *pflag.FlagSet, cfg *forwarder.TLSClientConfig) {
 		"Don't verify the server's certificate chain and host name. "+
 			"Enable to work with self-signed certificates. ")
 
-	fs.StringSliceVar(&cfg.CAFiles,
-		"cacert-file", cfg.CAFiles, "<path or base64>"+
+	fs.StringSliceVar(&cfg.CACertFiles,
+		"cacert-file", cfg.CACertFiles, "<path or base64>"+
 			"Add your own CA certificates to verify against. "+
 			"The system root certificates will be used in addition to any certificates in this list. "+
 			"Can be a path to a file or \"data:\" followed by base64 encoded certificate. "+
--- a/bind/flag.go
+++ b/bind/flag.go
@@ -105,6 +105,29 @@ func HTTPProxyConfig(fs *pflag.FlagSet, cfg *forwarder.HTTPProxyConfig, lcfg *lo
 			"By default, requests to localhost are denied. ")
 }
 
+func MITMConfig(fs *pflag.FlagSet, mitm *bool, cfg *forwarder.MITMConfig) {
+	fs.BoolVar(mitm, "mitm", *mitm, ""+
+		"Enable Man-in-the-Middle (MITM) mode. "+
+		"It only works with HTTPS requests, HTTP/2 is not supported. "+
+		"MITM is enabled by default when the --mitm-cacert-file flag is set. "+
+		"If the CA certificate is not provided MITM uses a generated CA certificate. "+
+		"The CA certificate used can be retrieved from the API server .")
+
+	fs.StringVar(&cfg.CACertFile, "mitm-cacert-file", cfg.CACertFile, "<path or base64>"+
+		"CA certificate file to use for generating MITM certificates. "+
+		"If the file is not specified, a generated CA certificate will be used. "+
+		"See the documentation for the --mitm flag for more details. ")
+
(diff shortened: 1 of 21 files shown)


bind/flag.go
```diff
@@ -105,6 +105,29 @@ func HTTPProxyConfig(fs *pflag.FlagSet, cfg *forwarder.HTTPProxyConfig, lcfg *lo
 			"By default, requests to localhost are denied. ")
 }
 
+func MITMConfig(fs *pflag.FlagSet, mitm *bool, cfg *forwarder.MITMConfig) {
+	fs.BoolVar(mitm, "mitm", *mitm, ""+
+		"Enable Man-in-the-Middle (MITM) mode. "+
+		"It only works with HTTPS requests, HTTP/2 is not supported. "+
+		"MITM is enabled by default when the --mitm-cacert-file flag is set. "+
+		"If the CA certificate is not provided MITM uses a generated CA certificate. "+
```
waggledans, line 113: > If the CA certificate is not provided MITM uses a generated CA certificate

I don't see a use case for that. I think we must require a cert when MITM is enabled

bind/flag.go
```diff
@@ -105,6 +105,29 @@ func HTTPProxyConfig(fs *pflag.FlagSet, cfg *forwarder.HTTPProxyConfig, lcfg *lo
...
+
+	fs.StringVar(&cfg.CACertFile, "mitm-cacert-file", cfg.CACertFile, "<path or base64>"+
+		"CA certificate file to use for generating MITM certificates. "+
+		"If the file is not specified, a generated CA certificate will be used. "+
+		"See the documentation for the --mitm flag for more details. ")
+
+	fs.StringVar(&cfg.CAKeyFile, "mitm-cakey-file", cfg.CAKeyFile, "<path or base64>"+
+		"CA key file to use for generating MITM certificates. ")
+
+	fs.StringVar(&cfg.Organization, "mitm-org", cfg.Organization, "<name>"+
```
waggledans, line 124: same comment as above. I would simplify the MITM functionality by requiring a cert to use for resigning. It doesn't seem right to let proxy generate certs (unless it's for testing).

bind/flag.go
```diff
@@ -105,6 +105,29 @@ func HTTPProxyConfig(fs *pflag.FlagSet, cfg *forwarder.HTTPProxyConfig, lcfg *lo
...
 
+func MITMConfig(fs *pflag.FlagSet, mitm *bool, cfg *forwarder.MITMConfig) {
+	fs.BoolVar(mitm, "mitm", *mitm, ""+
+		"Enable Man-in-the-Middle (MITM) mode. "+
+		"It only works with HTTPS requests, HTTP/2 is not supported. "+
+		"MITM is enabled by default when the --mitm-cacert-file flag is set. "+
+		"If the CA certificate is not provided MITM uses a generated CA certificate. "+
+		"The CA certificate used can be retrieved from the API server .")
+
+	fs.StringVar(&cfg.CACertFile, "mitm-cacert-file", cfg.CACertFile, "<path or base64>"+
```
Choraden, line 116: Cert and Key flags have `-file` suffix, which we previously removed from other cert flags. What's the reason behind this? Do we want to keep SC 4.X naming? 
IMO we must have the same naming convention for all cert related flags.

mitm.go
```diff
@@ -0,0 +1,67 @@
...
+	CAKeyFile  string
+
+	Organization string
+	Validity     time.Duration
+}
+
+func DefaultMITMConfig() *MITMConfig {
+	return &MITMConfig{
+		Organization: "Sauce Labs Inc.",
+		Validity:     24 * time.Hour, //nolint:gomnd // 24 hours is a reasonable default
```
Choraden, line 30: Let's add 24 to ignored-numbers in golangci file. As you explain, 24 is a reasonable value and occurs often when it comes to time.

bind/flag.go
mmatczuk: This is very useful for quick checks.
Then you can get the certificate from the API server and not bother generating it.
We already have sth very similar for running HTTPS with self-signed cert.

mmatczuk: Think Windows.

bind/flag.go
mmatczuk: See my other comment.

bind/flag.go
mmatczuk: We did not yet remove them. There is a separate ticket for it https://github.com/saucelabs/forwarder/issues/337.

mitm.go
mmatczuk: OK

mmatczuk: I'm disabling gomnd

PR #358, opened 2023-08-14
fix Excess found: excess = 2 url = / (zero-length body) in HEAD response

mmatczuk:
We need e2e HEAD test:
* with and without body
* Transfer-Encoding: (none just content-length), chunked, identity


Choraden (approved):
LGTM

PR #366, opened 2023-08-17
deny domains

Choraden:
This patch add the functionality to specify which domains should be denied from proxing. 
Please read commit messages for more info.

### DIFF
--- a/bind/flag.go
+++ b/bind/flag.go
@@ -11,6 +11,7 @@ import (
 	"net/netip"
 	"net/url"
 	"os"
+	"regexp"
 	"strings"
 
 	"github.com/mmatczuk/anyflag"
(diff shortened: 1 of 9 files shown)


Choraden:
E2E tests need to be added.

amckenzie132 (approved):
LGTM

bind/flag.go
```diff
@@ -103,6 +104,20 @@ func HTTPProxyConfig(fs *pflag.FlagSet, cfg *forwarder.HTTPProxyConfig, lcfg *lo
...
 			"By default, requests to localhost are denied. ")
+
+	fs.Var(anyflag.NewSliceValue[*regexp.Regexp](cfg.DenyDomains, &cfg.DenyDomains, regexp.Compile),
+		"deny-domain", "<regexp>"+
+			"Deny requests to the specified domains. "+
+			"Use this flag multiple times to specify multiple deny domains. "+
+			"Use --deny-domain-exclude to exclude requests to certain domains from being denied.")
+
+	fs.Var(anyflag.NewSliceValue[*regexp.Regexp](cfg.DenyDomainsExclude, &cfg.DenyDomainsExclude, regexp.Compile),
+		"deny-domain-exclude", "<regexp>"+
```
waggledans, line 115: I'm wondering whether we should use the following naming convention:

-  `--always-deny` and `--always-allow` since, technically, regexps are not domains.
- `--always-direct` and `--always-proxy` for the tunnel domains (in case of forwarder, it doesn't make sense to me to use `--always-tunnel`

amckenzie132: * I'm not sure about this convention. This means we have `--always` before every flag, but there's isn't a `--never` counterpart. It also means we need two different and opposite values for each flag set, eg. `deny` and `allow`, or `direct` and `proxy`. For TLS bumping, would we use `bump` and `splice`?

* What I like about the `--name` and `--name-exclude` convention is that the order is clear (which is checked first), and the pair is ordered sequentially in the docs.

* Regexps are not domains, but the regexp acts on the domain (`req.URL.Hostname()`) (side note: do we want domains only or full url matches)?

amckenzie132: Regexps are not domains, but the regexp acts on the domain (as currently implemented). So having a name like `--deny-domains` makes sense.

Side note: do we want domains only or full url matches?

Choraden: Agree with @amckenzie132 on `--deny-domains`. I'm not 100% sure about `--deny-domain-exclude`, but I like that '-exclude' suffix implies it's the mirror flag.

Choraden: >do we want domains only or full url matches?

Do we have a use case? IMO it's an overkill.

waggledans: I'm less concerned with Forwarder flag names and more with SC flags. Overall, the whole discussion about the flags should be in the context of all domain flags to make sure we have the naming scheme that works for all of them:
- deny/allow
- direct/tunnel
- mitm/passthrough

waggledans: After some discussion, I'm OK with the current implementation. I'd suggest not implementing direct/tunnel matching for now and focusing on mitm/passthrough as the next step.

waggledans (approved):

PR #370, opened 2023-08-25
bind: Add JSON parameter to DescribeFlags

alexh-sauce:
This PR adds an option to describe flags in JSON format.

This is useful for exporting or recording config settings.

### DIFF
--- a/bind/describe.go
+++ b/bind/describe.go
@@ -1,0 +1,62 @@
+// Copyright 2023 Sauce Labs Inc. All rights reserved.
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+package bind
+
+import (
+	"encoding/json"
+	"fmt"
+	"sort"
+	"strings"
+
+	"github.com/spf13/pflag"
+)
+
+type DescribeFormat int
+
+const (
+	Plain DescribeFormat = iota
+	JSON
+)
+
+func DescribeFlags(fs *pflag.FlagSet, showHidden bool, format DescribeFormat) (string, error) {
+	args := make(map[string]any, fs.NFlag())
+	keys := make([]string, 0, fs.NFlag())
+
+	fs.VisitAll(func(flag *pflag.Flag) {
+		if flag.Name == "help" {
+			return
+		}
+
+		if flag.Hidden && !showHidden {
+			return
+		}
+
+		if flag.Value.Type() == "bool" {
+			args[flag.Name] = flag.Value
+		} else {
+			args[flag.Name] = strings.Trim(flag.Value.String(), "[]")
+		}
+
+		keys = append(keys, flag.Name)
+	})
+
+	sort.Strings(keys)
+
+	switch format {
+	case Plain:
+		var b strings.Builder
+		for _, name := range keys {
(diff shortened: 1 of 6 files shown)


bind/flag.go
```diff
@@ -25,6 +27,13 @@ import (
 	"github.com/spf13/pflag"
 )
 
+type Format int
```
Choraden, line 30: Let's name it `DescribeFormat`. This is `bind` package and `bind.Format` might be misleading.

bind/flag.go
```diff
@@ -25,6 +27,13 @@ import (
...
+
+const (
+	Plain Format = 0
+	JSON  Format = 1
+)
```
Choraden, line 35: nit: 
```go
const (
	Plain Format = iota
	JSON
)
```

bind/flag.go
```diff
@@ -328,13 +337,41 @@ func MarkFlagFilename(cmd *cobra.Command, names ...string) {
...
+		if flag.Name == "help" {
+			return
+		}
+
+		if flag.Hidden && !showHidden {
 			return
 		}
-		b.WriteString(fmt.Sprintf("%s=%s\n", flag.Name, strings.Trim(flag.Value.String(), "[]")))
+
+		if flag.Value.Type() == "bool" {
```
Choraden, line 353: I don't get this case. Please explain.

Choraden: ah we don't want the booleans to be quoted.

bind/flag.go
```diff
@@ -328,13 +337,41 @@ func MarkFlagFilename(cmd *cobra.Command, names ...string) {
 	}
 }
 
-func DescribeFlags(fs *pflag.FlagSet) string {
-	var b strings.Builder
+func DescribeFlags(fs *pflag.FlagSet, showHidden bool, format Format) (string, error) {
+	args := make(map[string]any, 0)
+	keys := make([]string, 0)
```
Choraden, line 342: nit: use fs.NFlag() to preallocate

cmd/forwarder/run/run.go
```diff
@@ -43,7 +43,10 @@ type command struct {
 }
 
 func (c *command) RunE(cmd *cobra.Command, _ []string) error {
-	config := bind.DescribeFlags(cmd.Flags())
+	config, err := bind.DescribeFlags(cmd.Flags(), false, bind.JSON)
```
Choraden, line 46: How about we stick to the Plain format here?

Choraden:
How about we move it to a separate file e.g. `describe.go`?

cmd/forwarder/run/run.go
alexh-sauce: Oof, yes. Thanks for catching that.

alexh-sauce:
@Choraden Make sense, I've moved the function to a new file.

bind/describe.go
```diff
@@ -0,0 +1,62 @@
...
+type DescribeFormat int
+
+const (
+	Plain DescribeFormat = iota
+	JSON
+)
+
+func DescribeFlags(fs *pflag.FlagSet, showHidden bool, format DescribeFormat) (string, error) {
+	args := make(map[string]any, fs.NFlag())
+	keys := make([]string, fs.NFlag())
```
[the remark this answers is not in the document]
alexh-sauce: Ah, gotcha.

Choraden (approved):
LGTM

PR #372, opened 2023-08-28
Add route mode for http logging 

Choraden:
A `route` mode is like `url`, but it logs only the host+path to make the logs look thin.

Route log example: 
```
2023/08/28 16:54:40 [proxy] INFO: trace=2 GET www.google.com/ status=200 duration=3.140999458s
```

### DIFF
--- a/httplog/httplog.go
+++ b/httplog/httplog.go
@@ -21,6 +21,7 @@ type Mode string
 
 const (
 	None    Mode = "none"
+	Route   Mode = "route"
 	URL     Mode = "url"
 	Headers Mode = "headers"
 	Body    Mode = "body"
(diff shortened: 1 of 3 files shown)


httplog/httplog.go
```diff
@@ -48,25 +49,31 @@ func (l *Logger) LogFunc() middleware.Logger {
...
+			l.log("%s", w.String())
 		}
 	case Headers:
 		return func(e middleware.LogEntry) {
 			var w logWriter
-			w.Line(e)
+			w.RouteLine(e)
 			w.Dump(e)
-			l.log(w.String())
+			l.log("%s", w.String())
```
alexh-sauce, line 69: I'm curious why this line changed - will it now be lazily evaluated?

Choraden: It's like calling `fmt.Printf("%s", "text")` vs `fmt.Printf("text")`

waggledans (approved):

httplog/httplog.go
alexh-sauce: Oh, I see! A subtle bug, good catch.

PR #380, opened 2023-08-31
http_server: Add `BindPort` function

alexh-sauce:
In some cases, we want to check if the HTTP server can start successfully before moving on to initializing other resources.

This allows us to return an error early if the port isn't available.


http_server.go
```diff
@@ -189,6 +189,15 @@ func (hs *HTTPServer) configureHTTP2() error {
...
 }
 
+func (hs *HTTPServer) BindPort() error {
+	if hs.Listener != nil {
+		return fmt.Errorf("HTTP server already bound to port")
+	}
+
+	_, err := hs.listener()
+	return err
+}
```
Choraden, line 199: I believe, it's not going to work as we neither release the port, nor set `hs.Listener`.

Choraden: I see two approaches there:
1. ListenerDryRun -  try to bind and release the port.
2. Manual listener start - like you did but set hs.Listener.

alexh-sauce: Ugh, the commit I meant to push included 9735029

alexh-sauce: Binding and releasing the port won't work as it may use `0` to bind a random port, and then we introduce a race condition when reusing.

waggledans: I'm not sure I understand the problem. Why do we need this?
If the server fails to bind, we fail, no need to check for it, in advance

http_server.go
```diff
@@ -189,6 +189,16 @@ func (hs *HTTPServer) configureHTTP2() error {
 	return hs.config.ConfigureTLSConfig(hs.srv.TLSConfig)
 }
 
+func (hs *HTTPServer) BindPort() error {
```
Choraden, line 192: I would prefer a name closer to `RunListener`. WDYT?

Choraden: We could also add a comment that caller must close listener.

Choraden:
Please squash the commits.

Choraden:
After thinking about it, I doubt if we really need this. 

Adding `BindPort` definitely increases the code complexity there and puts us into handling different cases. 
 
The `Listener` is exported on purpose, so that users can specify it - let's use it.

alexh-sauce:
> The Listener is exported on purpose, so that users can specify it - let's use it.

Right, ok, that sounds good.

PR #381, opened 2023-09-01
utils/cobrautil: support pasring lists from env vars and config file

Choraden:
This trims and reformats string values set from viper. It allows reading yaml and json format lists e.g.
```
dns-server:
 - 1.2.3.4
 - 4.3.2.1
```
```
{
  "dns-server": ["8.8.8.8", "5.5.5.5"]
}
```
and env vars:
```
FORWARDER_DNS_SERVER="1.2.3.4 4.3.2.1" forwarder run
```

It does not affect reading and parsing flags from command line.

### DIFF
--- a/utils/cobrautil/bind.go
+++ b/utils/cobrautil/bind.go
@@ -55,7 +55,11 @@ func BindAll(cmd *cobra.Command, envPrefix, configFileFlagName string) error {
 		ok = true
 		fs.VisitAll(func(f *pflag.Flag) {
 			if !f.Changed && v.IsSet(f.Name) {
-				if err := fs.Set(f.Name, fmt.Sprintf("%v", v.Get(f.Name))); err != nil {
+				s := fmt.Sprintf("%v", v.Get(f.Name))
+				s = strings.TrimPrefix(s, "[")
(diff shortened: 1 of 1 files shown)


utils/cobrautil/bind.go
```diff
@@ -55,7 +55,11 @@ func BindAll(cmd *cobra.Command, envPrefix, configFileFlagName string) error {
 		ok = true
 		fs.VisitAll(func(f *pflag.Flag) {
 			if !f.Changed && v.IsSet(f.Name) {
-				if err := fs.Set(f.Name, fmt.Sprintf("%v", v.Get(f.Name))); err != nil {
+				s := fmt.Sprintf("%v", v.Get(f.Name))
+				s = strings.TrimPrefix(s, "[")
```
waggledans, line 59: Is it because we don't have access to the anyflag repo right now? I think it could wait a few days (unless it's problematic to fix in anyflag)

Choraden: After investigation, it turns out that anyflag follows parsing rules of cobra, so we probably don't want to mess with it. 
More advanced parsing is usually implemented in a higher layer and later passed to cobra like I've done here.

PR #388, opened 2023-09-04
add `/headers` endpoint to httpbin

Choraden:
This adds `/headers` endpoint to httpbin, which returns the incoming request's HTTTP headers.

The test `TestProxyUpstream` has been adjusted to use `/headers`.


mmatczuk:
LGTM.

You could also remove the `/header` endpoint as not part of the spec and takes the validation responsibility from the test wich results in double validation (to provide status code and the validation of the status code itself).

PR #396, opened 2023-09-06
refactor recent changes to martian logging

### LINKED ISSUE #394: refactor recent changes to martian logging
`martian/context: use base 10 when converting id to string Hubert Grochowski 21/08/2023, 19:09` - Drop it - it makes IDs longer, I doubt it adds value if numbers are high. Any calculator can count the offset in base 16.

`martian/context: store id as string Hubert Grochowski 21/08/2023, 19:11` - Drop it, it takes more memory and additional allocation on the heap and adds no value.

`martian/log: replace with forwarder logger Hubert Grochowski 21/08/2023, 16:33` - This is wrong, there should be no dependency on forwarder here. log should specify Logger interface that is implemented by forwarder logger.

An attractive alternative Martian logger integration design is by adding context to log functions and passing the logger in the context.

That means:
- all `log` functions get context.Context as the 1st parameter
- the `currLogger` is becomes `defaultLogger`
- Logger is added to Proxy
- when proxy reads request and creates context the logger is put to the context - this is a logger with request ID key/value pair
- for 3rd party code (martian sub packages) we use `context.TODO` which uses the `defaultLogger`
- for martian code we use request context

`log: extend Logger interface with WithValue method Hubert Grochowski 28/08/2023, 10:04` - could be squashed with the other commit where you add implementation.

`log/stdlog: rename StdLogger to Logger Hubert Grochowski 22/08/2023, 12:34` - OK, strange to find it in the middle of the patchset 

`log: use any instead of empty interface Hubert Grochowski 24/08/2023, 16:43` - OK

`maritan/log: implement WithTraceID Hubert Grochowski 22/08/2023, 12:14` - Drop it, not needed with proposal

`martian: use custom type for context key to avoid collision Hubert Grochowski 24/08/2023, 15:59` - We need the same for the logger

`martian: add function to retrieve martian.Conetext from context.Context Hubert Grochowski 24/08/2023, 16:03` - Drop it, not needed with proposal

`martian(context): do not initialize ID value Hubert Grochowski 22/08/2023, 14:35` - Drop it, this allows to work with IDs when the server restarts, otherwise you always start from 0. 

`martian(proxy): use traced logger Hubert Grochowski 28/08/2023, 10:57` - Uhggggg, drop it 

`httplog: log trace id Hubert Grochowski 28/08/2023, 11:02` - `trace=%s` ? how about `id=%d` 

`martian/messageview: redact url in request snapshot Hubert Grochowski 28/08/2023, 17:58` - OK, irrelevant in this patchset

`httplog: use string format for the logWriter content Hubert Grochowski 28/08/2023, 18:22` - this actually fixes a bug we had #300 I think I may have accidentally reference wrong issue.

Choraden:
Fixes #394 

Example log message:
```
[proxy] [DEBUG] [18a6b0c67a1] msg
```

### DIFF
--- a/internal/martian/log/log.go
+++ b/internal/martian/log/log.go
@@ -15,6 +15,11 @@
 // Package log provides a universal logger for martian packages.
 package log
 
+import (
+	"context"
+	"fmt"
+)
+
 type Logger interface {
 	Infof(format string, args ...any)
 	Debugf(format string, args ...any)
@@ -30,19 +35,30 @@ func SetLogger(l Logger) {
 	currLogger = l
 }
 
+type contextKey string
+
+const TraceContextKey contextKey = "trace"
+
 // Infof logs an info message.
-func Infof(format string, args ...any) {
-	currLogger.Infof(format, args...)
+func Infof(ctx context.Context, format string, args ...any) {
+	currLogger.Infof(withTrace(ctx, format), args...)
 }
 
 // Debugf logs a debug message.
-func Debugf(format string, args ...any) {
-	currLogger.Debugf(format, args...)
+func Debugf(ctx context.Context, format string, args ...any) {
+	currLogger.Debugf(withTrace(ctx, format), args...)
 }
 
 // Errorf logs an error message.
-func Errorf(format string, args ...any) {
-	currLogger.Errorf(format, args...)
+func Errorf(ctx context.Context, format string, args ...any) {
+	currLogger.Errorf(withTrace(ctx, format), args...)
+}
+
+func withTrace(ctx context.Context, format string) string {
+	if v := ctx.Value(TraceContextKey); v != nil {
+		format = fmt.Sprintf("[%s] %s", v, format)
+	}
+	return format
 }
 
 type nopLogger struct{}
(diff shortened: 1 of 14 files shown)


internal/martian/log/log.go
```diff
@@ -30,18 +35,31 @@ func SetLogger(l Logger) {
...
+type contextKey string
+
+const TraceContextKey contextKey = "trace"
+
 // Infof logs an info message.
-func Infof(format string, args ...any) {
+func Infof(ctx context.Context, format string, args ...any) {
+	if v := ctx.Value(TraceContextKey); v != nil {
+		format = fmt.Sprintf("[%s] %s", v, format)
+	}
```
mmatczuk, line 46: This is repeated 3 times, maybe extract that to function?

mmatczuk:
This probably can be squashed

martian(handler): use context logging functions Hubert Grochowski Yesterday 16:30
martian(proxy): use context logging functions Hubert Grochowski Yesterday 16:29

mmatczuk:
I suggest to move the cutting of martian log prefix to martian log functions instead of doing 

cmd/forarder/run: apply more advanced logic to cut `martian:` prefix from martian logger. Hubert Grochowski Yesterday 17:16

internal/martian/log/log.go
Choraden: ok

PR #401, opened 2023-09-07
utils/cobrautil(bind): set flags directly in FlagSet

Choraden:
pflag.FlagSet contains various metadata about the flags e.g. if one has been changed. Using pflag.Flag.Value.Set does not update pflag.FlagSet state.

The flags must be set by FlagSet.Set, otherwise it is easy to violate its internal state.

### DIFF
--- a/utils/cobrautil/bind.go
+++ b/utils/cobrautil/bind.go
@@ -76,7 +76,7 @@ func BindAll(cmd *cobra.Command, envPrefix, configFileFlagName string) error {
 }
 
 func setFlagFromViper(f *pflag.Flag, v any) error {
-	if vs, ok := v.([]interface{}); ok {
+	if vs, ok := v.([]any); ok {
 		sr, ok := f.Value.(sliceReplacer)
 		if !ok {
 			return fmt.Errorf("trying to set list to %s", f.Value.Type())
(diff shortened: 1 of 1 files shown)


waggledans (approved):

mmatczuk:
This commit butchers solution for no good reason.

So what exactly is the problem?

The code
* get flag
* call Set
* handle error `return fmt.Errorf("invalid argument %q for %q flag: %v", value, flagName, err)` - this we should add
* handle case where the value is not changed even after set - I guess we can return error here too
* print deprecation notice - maybe we should be adding it too

```
// Set sets the value of the named flag.
func (f *FlagSet) Set(name, value string) error {
	normalName := f.normalizeFlagName(name)
	flag, ok := f.formal[normalName]
	if !ok {
		return fmt.Errorf("no such flag -%v", name)
	}

	err := flag.Value.Set(value)
	if err != nil {
		var flagName string
		if flag.Shorthand != "" && flag.ShorthandDeprecated == "" {
			flagName = fmt.Sprintf("-%s, --%s", flag.Shorthand, flag.Name)
		} else {
			flagName = fmt.Sprintf("--%s", flag.Name)
		}
		return fmt.Errorf("invalid argument %q for %q flag: %v", value, flagName, err)
	}

	if !flag.Changed {
		if f.actual == nil {
			f.actual = make(map[NormalizedName]*Flag)
		}
		f.actual[normalName] = flag
		f.orderedActual = append(f.orderedActual, flag)

		flag.Changed = true
	}

	if flag.Deprecated != "" {
		fmt.Fprintf(f.out(), "Flag --%s has been deprecated, %s\n", flag.Name, flag.Deprecated)
	}
	return nil
}
```

mmatczuk:
LGTM

PR #403, opened 2023-09-08
add `--proxy-header` for managing proxy headers

Choraden:
This allows to modify CONNECT request to upstream proxy.

### DIFF
--- a/dialvia/http.go
+++ b/dialvia/http.go
@@ -22,6 +22,8 @@ type HTTPProxyDialer struct {
 	dial      ContextDialerFunc
 	proxyURL  *url.URL
 	tlsConfig *tls.Config
+
+	ConnectRequestModifier func(req *http.Request) error
 }
 
 func HTTPProxy(dial ContextDialerFunc, proxyURL *url.URL) *HTTPProxyDialer {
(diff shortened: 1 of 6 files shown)


dialvia/http.go
```diff
@@ -19,9 +19,10 @@ import (
 )
 
 type HTTPProxyDialer struct {
-	dial      ContextDialerFunc
-	proxyURL  *url.URL
-	tlsConfig *tls.Config
+	dial                   ContextDialerFunc
+	proxyURL               *url.URL
+	tlsConfig              *tls.Config
+	ConnectRequestModifier func(req *http.Request) error
```
mmatczuk, line 25: I guess it would be easier if it could not return error.

Think [Rewrite](https://cs.opensource.google/go/go/+/refs/tags/go1.21.1:src/net/http/httputil/reverseproxy.go;l=125).

Please add to separate group

dialvia/http_test.go
```diff
@@ -134,6 +135,58 @@ func TestHTTPProxyDialerDialContext(t *testing.T) {
 	})
 }
 
+func TestHTTPProxyConnectRequestModifier(t *testing.T) {
```
mmatczuk, line 138: I love tests but I guess this one adds no value tbh.

dialvia/http.go
mmatczuk: After seeing it wider context it makes sense.

dialvia/http_test.go
Choraden: OK I can remove it, since the implementation is short and straightforward. Moreover, there is another test of that in martian.

dialvia/http.go
Choraden: I moved it to a separate group.

PR #405, opened 2023-09-08
Eager listener init

mmatczuk:


Choraden:
I don't like that there is more the one subject which closes the listener i.e.`Proxy.Close` and `Proxy.Run`, but it's not wrong though, just my personal feeling.

Choraden:
How about documenting that one needs to close the Proxy once created?

mmatczuk:
> I don't like that there is more the one subject which closes the listener i.e.Proxy.Close and Proxy.Run, but it's not wrong though, just my personal feeling.

It's needed in case of martian so that it stop listening, it is indeed wrong.

mmatczuk:
Added comments.

Choraden:
I'm wondering if we could add `Proxy::close()` and call it in both `Proxy:Close()`, `Proxy:Run()` to indicate that they both do the same.

mmatczuk:
Done.

PR #407, opened 2023-09-11
Read request ID from header and use in logs

Choraden:
The proxy is capable to associate the request id send in request header with logs.

### DIFF
--- a/bind/flag.go
+++ b/bind/flag.go
@@ -127,6 +127,11 @@ func HTTPProxyConfig(fs *pflag.FlagSet, cfg *forwarder.HTTPProxyConfig, lcfg *lo
 			"Use this flag multiple times to specify multiple deny domain excludes. "+
 			"This flag takes precedence over deny-domain. "+
 			"Can be specified only if --deny-domain is also specified. ")
+
+	fs.StringVar(&cfg.RequestIDHeader, "log-http-request-id-header", cfg.RequestIDHeader,
+		"<name>"+
+			"If the header is present in the request, "+
+			"the proxy will associate the value with the request in the logs. ")
 }
 
 func MITMConfig(fs *pflag.FlagSet, mitm *bool, cfg *forwarder.MITMConfig) {
--- a/http_proxy.go
+++ b/http_proxy.go
@@ -79,6 +79,7 @@ type HTTPProxyConfig struct {
 	ProxyLocalhost         ProxyLocalhostMode
 	UpstreamProxy          *url.URL
 	UpstreamProxyFunc      ProxyFunc
+	RequestIDHeader        string
 	RequestModifiers       []RequestModifier
 	ResponseModifiers      []ResponseModifier
 	ConnectRequestModifier func(*http.Request) error
@@ -100,8 +101,9 @@ func DefaultHTTPProxyConfig() *HTTPProxyConfig {
 			ReadHeaderTimeout: 1 * time.Minute,
 			LogHTTPMode:       httplog.Errors,
 		},
-		Name:           "forwarder",
-		ProxyLocalhost: DenyProxyLocalhost,
+		Name:            "forwarder",
+		ProxyLocalhost:  DenyProxyLocalhost,
+		RequestIDHeader: "X-Request-Id",
 	}
 }
 
@@ -225,6 +227,7 @@ func (hp *HTTPProxy) configureProxy() error {
 	}
 
 	hp.proxy.AllowHTTP = true
+	hp.proxy.RequestIDHeader = hp.config.RequestIDHeader
 	hp.proxy.ConnectRequestModifier = hp.config.ConnectRequestModifier
 	hp.proxy.ConnectPassthrough = hp.config.ConnectPassthrough
 	hp.proxy.WithoutWarning = true
--- a/internal/martian/handler.go
+++ b/internal/martian/handler.go
@@ -89,7 +89,7 @@ func (p proxyHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 	}
 	ctx := withSession(session)
 
-	outreq := req.Clone(ctx.addToContext(req.Context()))
+	outreq := req.Clone(p.requestContext(ctx, req))
 	if req.ContentLength == 0 {
 		outreq.Body = http.NoBody
 	}
--- a/internal/martian/proxy.go
+++ b/internal/martian/proxy.go
@@ -62,6 +62,11 @@ type Proxy struct {
 	// AllowHTTP disables automatic HTTP to HTTPS upgrades when the listener is TLS.
 	AllowHTTP bool
 
+	// RequestIDHeader specifies a special header name that the proxy will use to identify requests.
+	// If the header is present in the request, the proxy will associate the value with the request in the logs.
+	// If empty, no action is taken, and the proxy will generate a new request ID.
+	RequestIDHeader string
+
 	// ConnectPassthrough passes CONNECT requests to the RoundTripper,
 	// and uses the response body as the connection.
 	ConnectPassthrough bool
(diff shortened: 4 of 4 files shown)


internal/martian/handler.go
```diff
@@ -89,7 +89,11 @@ func (p proxyHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 	}
 	ctx := withSession(session)
 
-	outreq := req.Clone(ctx.addToContext(req.Context()))
+	cctx := ctx.addToContext(req.Context())
```
mmatczuk, line 92: How about `outreq` modification if needed?

internal/martian/proxy.go
```diff
@@ -368,7 +368,11 @@ func (p *Proxy) readRequest(ctx *Context, conn net.Conn, brw *bufio.ReadWriter)
 		default:
 		}
 
-		req = req.WithContext(ctx.addToContext(req.Context()))
+		cctx := ctx.addToContext(req.Context())
```
mmatczuk, line 371: Same here.

internal/martian/header/reqid_modifier.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/saucelabs/forwarder/internal/martian"
+	"github.com/saucelabs/forwarder/internal/martian/log"
+)
+
+// NewRequestIDModifier adds X-Request-ID header to the request if it is not already present.
+func NewRequestIDModifier() martian.RequestModifierFunc {
```
mmatczuk, line 12: 👍

internal/martian/httpspec/httpspec.go
```diff
@@ -33,6 +33,7 @@ func NewStack(via string) (outer, inner *fifo.Group) {
 	outer.AddRequestModifier(hbhm)
 	outer.AddRequestModifier(header.NewForwardedModifier())
 	outer.AddRequestModifier(header.NewBadFramingModifier())
+	outer.AddRequestModifier(header.NewRequestIDModifier())
```
mmatczuk, line 36: This is not part of `httpspec`!

Actually this should be controlled by a flag and maybe users should have control over name because adding X-Request-ID may cause some unexpected behavior.

internal/martian/proxy.go
```diff
@@ -822,7 +822,20 @@ func (p *Proxy) connectHTTP(req *http.Request, proxyURL *url.URL) (res *http.Res
 	} else {
 		d = dialvia.HTTPProxy(p.dial, proxyURL)
 	}
-	d.ConnectRequestModifier = p.ConnectRequestModifier
+	d.ConnectRequestModifier = func(r *http.Request) error {
```
mmatczuk, line 825: This does not belong here.

internal/martian/handler.go
Choraden: I did not want to make another copy of a request, but I can change it.

internal/martian/httpspec/httpspec.go
Choraden: You're right, it's not part of the spec. It was comfortable to put it here.

Choraden: I will add a flag.

internal/martian/proxy.go
Choraden: I know, but you need to somehow get the trace ID and CONNECT request won't have it in its context. That's why it's here.

internal/martian/handler.go
mmatczuk: Maybe refactor context creation to a function?

internal/martian/proxy.go
mmatczuk: I think it does not belong to Marian but Forwarder and should be implemented there.
There should be zero changes to martian code.

Choraden:
I added `add-request-id` flag that controls whether to pass trace id as X-Request-ID. 

Unfortunately, we can't reuse X-Request-ID from Forwarder, as it requires modifying request context - we would need to overwrite a pointer. 

I also dropped support for passing X-Request-ID in CONNECT requests, as it is infeasible outside `martian.Proxy`.

internal/martian/proxy.go
```diff
@@ -836,7 +850,19 @@ func (p *Proxy) connectHTTP(req *http.Request, proxyURL *url.URL) (res *http.Res
...
```
mmatczuk, line 853: Maybe we should have it as a function `p.connectRequestModifier` WDYT?

bind/flag.go
```diff
@@ -127,6 +127,15 @@ func HTTPProxyConfig(fs *pflag.FlagSet, cfg *forwarder.HTTPProxyConfig, lcfg *lo
...
 			"Can be specified only if --deny-domain is also specified. ")
+
+	fs.StringVar(&cfg.ReuseRequestIDFromHeader, "reuse-request-id-from-header", cfg.ReuseRequestIDFromHeader,
+		"<header>"+
+			"If the header is present in the request, "+
+			"the proxy will associate the value with the request in the logs. ")
+
+	fs.StringVar(&cfg.AddRequestIDHeader, "add-request-id-header", cfg.AddRequestIDHeader,
+		"<header>"+
+			"header name that the proxy will add to outgoing requests with id value. ")
```
mmatczuk, line 138:
```suggestion
The header name that the proxy will add to outgoing requests with internal request ID value.
```

internal/martian/proxy.go
```diff
@@ -104,6 +104,11 @@ type Proxy struct {
 	// CloseAfterReply closes the connection after the response has been sent.
 	CloseAfterReply bool
 
+	// ReuseRequestIDFromHeader specifies a special header name that the proxy will use to identify requests.
```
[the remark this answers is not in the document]
Choraden: ok

internal/martian/proxy.go
Choraden: The only problem is that we need to pass a context from the original request,  so it would be 
```go
p.connectRequestModifier(ctx context.Context)
```
but sure I can do that.

mmatczuk:
I have two thoughts strange naming and overall structure.

I'd probably want to see: 

`--log-http-request-id-header` with default value of "x-request-id" which means if request has this ID use it in logs - this is a logging feature

`--proxy-header x-request-id: {{req.ID()}}` - this is a syntax that treats header values like text template that can be evaluated against the request. ID then in a special function to get the request ID. This is much more flexible and elegant.

Choraden:
This PR will add only the feature of reading request ID and using it in logs.

internal/martian/proxy.go
```diff
@@ -62,6 +62,11 @@ type Proxy struct {
 	// AllowHTTP disables automatic HTTP to HTTPS upgrades when the listener is TLS.
 	AllowHTTP bool
 
+	// LogHTTPRequestIDHeader specifies a special header name that the proxy will use to identify requests.
+	// If the header is present in the request, the proxy will associate the value with the request in the logs.
+	// If empty, no action is taken, and the proxy will generate a new request ID.
+	LogHTTPRequestIDHeader string
```
mmatczuk, line 68: I think here it should be `RequestIDHeader`.

http_proxy.go
```diff
@@ -86,6 +86,7 @@ type HTTPProxyConfig struct {
 	CloseAfterReply        bool
 	DenyDomains            []*regexp.Regexp
 	DenyDomainsExclude     []*regexp.Regexp
+	LogHTTPRequestIDHeader string
```
mmatczuk, line 89: I do not like the practice of adding to the end of the list in random moments it does not make sense.

I think here it should also be named RequestIDHeader and live above RequestModifiers. WDYT?

mmatczuk:
[http_proxy: enable LogHTTPRequestIDHeader](https://github.com/saucelabs/forwarder/pull/407/commits/7107591116ffec179731b5a60664b3937afa8dd3) and the last commit can be squashed.

http_proxy.go
Choraden: Will do it in #414

PR #408, opened 2023-09-11
Allow getting inner proxy

mmatczuk:
This patch removes Tunnel scheme and proxy Handler getter.
It is replaced by a new constructor that returns a configured Martian proxy.


Choraden (approved):
LGTM

PR #411, opened 2023-09-11
Add rate limiting 

Choraden:
This adds global read and write rate limit.


http_proxy.go
```diff
@@ -86,6 +87,8 @@ type HTTPProxyConfig struct {
 	CloseAfterReply        bool
 	DenyDomains            []*regexp.Regexp
 	DenyDomainsExclude     []*regexp.Regexp
+	ReadLimit              uint
```
mmatczuk, line 90: I think it would be beneficial to use a special type that supports kb mb etc, see in rclone for inspiration.

bind/flag.go
```diff
@@ -127,6 +127,12 @@ func HTTPProxyConfig(fs *pflag.FlagSet, cfg *forwarder.HTTPProxyConfig, lcfg *lo
 			"Use this flag multiple times to specify multiple deny domain excludes. "+
 			"This flag takes precedence over deny-domain. "+
 			"Can be specified only if --deny-domain is also specified. ")
+
+	fs.UintVar(&cfg.ReadLimit, "read-limit", cfg.ReadLimit, "<bandwidth>"+
+		"Global read rate limit in bytes per second.")
```
mmatczuk, line 132: Same here more sophisticated input would be nice.

mmatczuk:
Looks good, some commits could be squashed.

resource/sizesuffix.go
```diff
@@ -0,0 +1,247 @@
+package resource
```
mmatczuk, line 1: How about putting it directly in forwarder. This is needed for config and CLI.

ratelimit/ratelimit.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"github.com/saucelabs/forwarder/resource"
+	"golang.org/x/time/rate"
+)
+
+const defaultMaxBurstSize = 4 * 1024 * 1024 // Must be bigger than the biggest request.
+
+func newRateLimiter(bandwidth resource.SizeSuffix) *rate.Limiter {
+	// The rate limiter is configured to allow a burst of 1/64th of the bandwidth or defaultMaxBurstSize.
```
mmatczuk, line 17: Let's leave the original discussion and a pointer to rclone issue.
The point is that it does not allow high bandwidth if defaultMaxBurstSize is used.

resource/sizesuffix.go
Choraden: ok

e2e/tests/flag_test.go
```diff
@@ -94,3 +97,50 @@ func TestFlagDenyDomain(t *testing.T) {
 	newClient(t, "https://www.google.com").GET("/").ExpectStatus(http.StatusForbidden)
 	newClient(t, httpbin).GET("/status/200").ExpectStatus(http.StatusOK)
 }
+
```
mmatczuk, line 100: Those tests do not make any sense to me.
* There should be one setup with rx and tx
* Test must use more than one connection
*  TestFlagWriteLimit is in fact read limit
* TestFlagReadLimit is doing sth very strange

We need an httpbin endpoint that copies request body to io.Discard

utils/httpbin/httpbin.go
```diff
@@ -24,6 +24,7 @@ func Handler() http.Handler {
 	m.HandleFunc("/delay/", delayHandler)
 	m.HandleFunc("/status/", statusHandler)
 	m.HandleFunc("/stream-bytes/", streamBytesHandler)
+	m.HandleFunc("/stream-discard/", streamDiscardHandler)
```
mmatczuk, line 27: stream comes from [Dynamic data](https://httpbin.org/#/Dynamic%20data) Generates random and dynamic data

It's more like sink just read all body, it feels like any function should do it.
You could be more creative and have count-bytes that one would return body size as a header etc.

e2e/setups.go
```diff
@@ -359,6 +360,45 @@ func SetupFlagDenyDomain(l *setupList) {
 	)
 }
 
+func SetupFlagRateLimit(l *setupList) {
+	l.Add(
```
mmatczuk, line 364: How about a single setup?

e2e/tests/flag_test.go
```diff
@@ -94,3 +98,77 @@ func TestFlagDenyDomain(t *testing.T) {
...
+	for i := 0; i < connections; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			c := newClient(t, httpbin)
+			c.GET(fmt.Sprintf("/stream-bytes/%d", size)).ExpectStatus(http.StatusOK).ExpectBodySize(size)
+		}()
+	}
+	wg.Wait()
+	if elapsed := time.Since(ts); elapsed.Round(time.Second) != expectedTime {
```
mmatczuk, line 121: Rounding is not great, let's say it can be 5% off and compute.

e2e/tests/flag_test.go
```diff
@@ -94,3 +98,79 @@ func TestFlagDenyDomain(t *testing.T) {
...
 	newClient(t, httpbin).GET("/status/200").ExpectStatus(http.StatusOK)
 }
+
+func TestFlagReadLimit(t *testing.T) {
+	const (
+		// It streams 2 * 5MiB, the read limit is 1MiB/s, but minimum burst is 4MiB, so it should take approximately 6 seconds.
+		connections     = 2
+		size            = 5 * 1024 * 1024 // 5MiB
+		expectedTime    = 6 * time.Second
+		acceptableRange = expectedTime * 5 / 100 // 5%
```
mmatczuk, line 108: how about epsilon.

e2e/tests/flag_test.go
```diff
@@ -94,3 +98,79 @@ func TestFlagDenyDomain(t *testing.T) {
...
+			}).ExpectStatus(http.StatusOK).ExpectHeader("body-size", fmt.Sprintf("%d", size))
+		}()
+	}
+	wg.Wait()
+	if elapsed := time.Since(ts); elapsed < expectedTime-acceptableRange || elapsed > expectedTime+acceptableRange {
+		t.Fatalf("Expected request to take approximately %s, took %s", expectedTime, elapsed)
+	}
+}
+
+func TestFlagReadWriteLimit(t *testing.T) {
```
mmatczuk, line 155: I do not see why we would want to run it twice.

Choraden: What do we run twice? It runs Read and Write test concurrently.

e2e/tests/flag_test.go
Choraden: Sounds good.

PR #419, opened 2023-09-13
use `-` syntax to exclude deny domain 

Choraden:


ruleset/ruleset.go
```diff
@@ -53,3 +53,24 @@ func (r *Regexp) Match(s string) bool {
 	}
 	return r.include != nil && r.include.MatchString(s)
 }
+
+// ParseRegexp parses the given rules into a Regexp.
+// Rules that start with "-" are treated as exclude rules.
+func ParseRegexp(in []string) (*Regexp, error) {
```
mmatczuk, line 59: Test would be nice.

http_proxy.go
```diff
@@ -85,8 +84,7 @@ type HTTPProxyConfig struct {
 	ConnectRequestModifier func(*http.Request) error
 	ConnectPassthrough     bool
 	CloseAfterReply        bool
-	DenyDomains            []*regexp.Regexp
-	DenyDomainsExclude     []*regexp.Regexp
+	DenyDomains            []string
```
mmatczuk, line 87: We could use the ruleset here to avoid double parsing.

mmatczuk:
It should be --deny-domains as you match many domains

mmatczuk:
Work moved to #423

PR #423, opened 2023-09-14
Deny domains new approach

mmatczuk:


ruleset/regexp.go
```diff
@@ -39,17 +39,47 @@ func NewRegexp(include, exclude []*regexp.Regexp) (*Regexp, error) {
...
 }
+
+type RegexpListItem struct {
+	*regexp.Regexp
+	exclude bool
+}
+
+func ParseRegexpListItem(v string) (RegexpListItem, error) {
+	var exclude bool
+	if v[0] == '-' {
```
Choraden, line 64: It would panic if string is empty. We should check it.

mmatczuk: true.

Choraden (approved):

PR #429, opened 2023-09-15
Support URL encoded username and password in basic authentication

Choraden:


Choraden:
By the way, I would like to point out that the forwarder passes the auth header to upstream proxies, while basic auth wrapper for http.Handler deletes it[1]. Is it intentional? 

[1] https://github.com/saucelabs/forwarder/blob/main/middleware/basic_auth.go#L103

config.go
```diff
@@ -33,11 +33,11 @@ func ParseUserInfo(val string) (*url.Userinfo, error) {
 	}
 
 	// URL decode the username and password
-	u, err := url.QueryUnescape(u)
+	u, err := url.PathUnescape(u)
```
mmatczuk, line 36: Interesting.

middleware/basic_auth_test.go
```diff
@@ -56,3 +56,16 @@ func TestBasicAuthWrap(t *testing.T) {
 		}
 	})
 }
+
+func TestBasicAuthURLEncoded(t *testing.T) {
```
mmatczuk, line 60: I do not like that you put things always at the end I causes more conflicts than needed and introduces balagan.
This should live below TestBasicAuth or better be a subtest.

Choraden: Well, if you think about, it lived below TestBasicAuth, just not directly... It's a subtest now :)

mmatczuk:
Closing this.

Issue #379, opened 2023-08-31
e2e: add tests for head requests 


Choraden:
We need e2e HEAD test:

- with and without body
- Transfer-Encoding: (none just content-length), chunked, identity

mmatczuk:
Writing test is difficult because it just prints to stderr a warning.
We may add a check that treats any stderr output as test error.
In that case we need to ensure that there are no stderr writes in the happy path.

PR #430, opened 2023-09-15
e2e: treat any stderr output as an error

Choraden:
This is to strengthen the tests and assert head requests are fine (#379)


mmatczuk:
ok

PR #433, opened 2023-09-19
Do not url escape user and password

Choraden:


credentials.go
```diff
@@ -10,14 +10,79 @@ import (
 	"fmt"
 	"net"
 	"net/url"
+	"strings"
 
 	"github.com/saucelabs/forwarder/log"
 )
 
+type Userinfo struct {
```
mmatczuk, line 18: It should be `UserInfo`

mmatczuk: let's move it to separate file.

credentials.go
```diff
@@ -10,14 +10,79 @@ import (
...
+	Username string
+	Password string
+}
+
+// ParseUserinfo parses a user:password string into Userinfo.
+// Username and password cannot be empty.
+func ParseUserinfo(val string) (*Userinfo, error) {
```
mmatczuk, line 25: Maybe we can change the API to use `Userinfo` not a pointer - this type is just two strings.
It would never escape and the API would not allow modifications by default.

credentials.go
mmatczuk, line 18: We should add String method that returns `<username>[:<password>]`.

mmatczuk:
We also have 

```
The server may be protected by basic authentication.
Whenever applicable, username and password are URL decoded.
This allows you to pass in special characters such as @ by using %%40 or pass in a colon with %%3a.
```

in `cmd`

config.go
```diff
@@ -105,11 +47,11 @@ func ParseProxyURL(val string) (*url.URL, error) {
 
 	creds, hostport, ok := strings.Cut(hostport, "@")
 	if ok {
-		ui, err := ParseUserInfo(creds)
+		ui, err := ParseUserinfo(creds)
 		if err != nil {
 			return nil, err
 		}
-		u.User = ui
+		u.User = url.UserPassword(ui.Username, ui.Password)
```
mmatczuk, line 54: I think this could be moved to Userinfo type like `AsURLUserinfo`, and the type needs some comment why it exists.

mmatczuk:
Let's update passwords in e2e to sth like `!@#$%^&*()_{};:'"?`

Choraden:
I removed the custom `Userinfo` type and changed the parsing functions to read unescaped user and pass.

config.go
```diff
@@ -21,28 +21,21 @@ import (
...
+func ParseUserInfo(in string) (*url.Userinfo, error) {
+	if in == "" {
+		return nil, fmt.Errorf("expected username[:password]")
 	}
-
-	u, p, ok := strings.Cut(val, ":")
-	if !ok {
-		return nil, fmt.Errorf("expected username:password")
+	if strings.Index(in, ":") != strings.LastIndex(in, ":") {
+		return nil, fmt.Errorf("only one colon is allowed")
```
mmatczuk, line 29: I think that can be potentially limiting.
The old logic says that username cannot have a colon which is more flexible I think.

mmatczuk:
The Parse function input parameter is often called `val` the naming is inherited from cobra.
I think that adding the rename makes the diff harder to read.

dialvia/http.go
```diff
@@ -118,8 +118,10 @@ func (d *HTTPProxyDialer) DialContextR(ctx context.Context, network, addr string
 
 	// Don't send the default Go HTTP client User-Agent.
 	req.Header.Add("User-Agent", "")
-	if d.proxyURL.User != nil {
-		req.Header.Add("Proxy-Authorization", "Basic "+base64.StdEncoding.EncodeToString([]byte(d.proxyURL.User.String())))
+	if u := d.proxyURL.User; u != nil {
+		pass, _ := u.Password()
+		auth := u.Username() + ":" + pass
```
mmatczuk, line 123: This is wrong. If password is empty it should be just `<username>`.

Choraden: Can change it, tho.

mmatczuk: That's actually a legit question if no password is the same as empty password.
I guess it depends.

Choraden: It is the same behaviour as for normal requests https://github.com/saucelabs/forwarder/blob/09bd2ad6723ef8f53bb8a98764309aa988e22147/http_proxy.go#L510

```go
func (hp *HTTPProxy) setBasicAuth(req *http.Request) error {
	if req.Header.Get("Authorization") == "" {
		if u := hp.creds.MatchURL(req.URL); u != nil {
			p, _ := u.Password()
			req.SetBasicAuth(u.Username(), p)
		}
	}

	return nil
}

func (r *Request) SetBasicAuth(username, password string) {
	r.Header.Set("Authorization", "Basic "+basicAuth(username, password))
}

func basicAuth(username, password string) string {
	auth := username + ":" + password
	return base64.StdEncoding.EncodeToString([]byte(auth))
}
```

I'm not sure if it is intended to send basic-auth without password.

config.go
Choraden: OK, removed this check.

PR #434, opened 2023-09-19
mitm domains

mmatczuk:


cmd/forwarder/run/run.go
```diff
@@ -154,8 +155,19 @@ func (c *command) runE(cmd *cobra.Command, _ []string) (cmdErr error) {
 	if c.mitmConfig.CACertFile != "" {
 		c.mitm = true
 	}
+	if len(c.mitmDomains) > 0 {
+		c.mitm = true
+	}
```
Choraden, line 160: How about we 'or' it with the condition above?

mmatczuk: Fixed that.

e2e/tests/flag_test.go
```diff
@@ -117,12 +117,19 @@ func TestFlagMITMDomains(t *testing.T) {
...
+func TestFlagDenyDomains(t *testing.T) {
+	t.Run("include(google)", func(t *testing.T) {
+		newClient(t, "https://www.google.com").GET("/").
+			ExpectStatus(http.StatusForbidden)
+	})
+
+	t.Run("exclude (httpbin)", func(t *testing.T) {
+		newClient(t, httpbin).GET("/status/200").
+			ExpectStatus(http.StatusOK)
+	})
```
Choraden, line 129: This change does not resonate with the commit name. 
Also make subtests names consistent - spaces.

Choraden:
LGTM. 

Good job with [martian (mitm): use Peek and avoid allocating reader for 1 byte slice](https://github.com/saucelabs/forwarder/pull/434/commits/45cb6201305b35f116f69de6db1c5dfebf0fe320) I really like this change.

e2e/tests/flag_test.go
mmatczuk: Spaces makes it look like `foo_(` vs `foo(`

mmatczuk: Updated the commit msg

mmatczuk:
Fixes #336

PR #440, opened 2023-09-21
header: support custom functions to evaluate add header value

Choraden:
This patch adds `id` function to declare more advanced add header value. 
Usage: `--proxy-header: "X-Request-Id: {{id}}"` - This will add a header with request's internal id.


mmatczuk:
We should approach it in 1.2 proper.

Choraden:
Closed as stale.

PR #443, opened 2023-09-26
middleware: remove metrics labels

amckenzie132:
Remove host and src metrics labels from http metrics to reduce cardinality

### DIFF
--- a/middleware/prometheus.go
+++ b/middleware/prometheus.go
@@ -8,7 +8,6 @@ package middleware
 
 import (
 	"fmt"
-	"net"
 	"net/http"
 	"strconv"
 	"time"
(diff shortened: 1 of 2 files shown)


alexh-sauce (approved):
LGTM

Choraden:
I doubt whether we want to remove the labels entirely. How about we make it configurable? 

Something like: `--http-metrics-labels` that takes a list of enabled labels. 
Does it make sense? Will the graphs on the dashboards autoupdate if the labels change?

mmatczuk (approved):

mmatczuk:
I think the issue is this inflates the metrics a lot @Choraden.

PR #447, opened 2023-09-27
api: custom api endpoints can overwrite defaults

amckenzie132:
Allow the default endpoints, `/metrics`, `/readyz`,`/healthz`, (and even`/debug/pprof`) to be overwritten in `customEndpoints`.


mmatczuk:
I do not think this is the right approach. We should have a default endpoints specified (like in cmd) and then anyone can change them.

PR #448, opened 2023-10-03
martian(mitm): check if peeked data is not nil

Choraden:
This check prevents the proxy from panic in the following scenario:
1. There isn't any buffered data.
2. Connection goes down.
3. Peek returns an error and nil slice.
4. Checking the first byte to determine a type results in reading from an empty slice.


mmatczuk:
Nice catch!

PR #463, opened 2023-10-12
runctx: allow to start group without notify signals

Choraden:


runctx/runctx.go
```diff
@@ -32,7 +32,8 @@ type Group struct {
 
 func NewGroup(fn ...func(ctx context.Context) error) *Group {
```
Choraden, line 33: I thinking of renaming it to `DefaultSignalsGroup`. WDYT @mmatczuk?

mmatczuk: Nah

mmatczuk (approved):
LGTM

PR #464, opened 2023-10-13
gh: run unit tests on Ubuntu and Windows

mmatczuk:
Fix Martian connection error handling on Windows and run unit tests on Windows runner.


runctx/runctx_unix_test.go
```diff
@@ -0,0 +1,30 @@
+// Copyright 2023 Sauce Labs Inc. All rights reserved.
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//go:build unix
```
Choraden, line 7: suggestion: `go:build unix` or `go:build !windows`

internal/martian/proxy.go
```diff
@@ -89,18 +89,23 @@ func isClosedConnError(err error) bool {
...
 
-	if errors.Is(err, io.EOF) ||
-		errors.Is(err, io.ErrClosedPipe) ||
-		errors.Is(err, errClose) {
+	if strings.Contains(err.Error(), "tls:") {
 		return true
 	}
 
-	return strings.Contains(err.Error(), "tls:")
+	return false
```
Choraden, line 108: Why so passive aggressive? :)

Let's just return `strings.Contains(err.Error(), "tls:")`

runctx/runctx_unix_test.go
mmatczuk: I do not get it? It is `//go:build unix`, `go:build !windows` does not apply here.

.golangci.yml
```diff
@@ -78,6 +78,7 @@ issues:
     - "Error return value of `cmd.MarkFlag.+` is not checked"
     - "Error return value of `net.SplitHostPort` is not checked"
     - "Error return value of `w.Write` is not checked"
+    - "Function 'handle' has too many statements"
```
Choraden, line 81: We keep silence these linters or increase limits. Maybe we should disable it for good.

runctx/runctx_unix_test.go
Choraden: I meant to use this implementation for all platforms that are not windows, not only unix.

Choraden: But maybe it's better to be straightforward here.

config_test.go
```diff
@@ -252,6 +253,10 @@ func TestParseFilePath(t *testing.T) {
 
 	for i := range tests {
 		tc := &tests[i]
+		if runtime.GOOS == "windows" {
+			tc.err = ""
+		}
```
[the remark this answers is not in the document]
mmatczuk: Added a comment.

Choraden (approved):

mmatczuk:
> I can see plenty of gh actions were run :)

Yes, as I tried to run all the tests like we do in other places.
To do that I'd need to get a local Windows access first.

.golangci.yml
mmatczuk: OK.

PR #467, opened 2023-10-17
cmd/forwarder/run: register process metrics

mmatczuk:
Since we are using a custom registry we need to add Go and process stats. Unfortunately the process data is only available in Linux and Windows.


Choraden (approved):

mmatczuk:
cc @amckenzie132

Issue #376, opened 2023-08-30
Question: Lots of errors "wsarecv: An existing connection was forcibly closed by the remote host."


bvarnai:
I have a simple setup, http proxy with pac, only browser agents (Firefox 116.0.3 mainly). I have VPN, direct and upstream proxies are on VPN.

`forwarder run --pac "http://proxyconf/" -H "User-Agent: Mozilla/5.0 (Windows NT 10.0
; Win64; x64)"`

I keep getting tons of errors (like it hogs the machine) on Windows 11, x64, version 1.0.2

> 2023/08/30 03:28:04 [proxy] ERROR: failed to read request: read tcp 172.30.235.203:3128->139.181.198.73:43028: wsarecv: An existing connection was forcibly closed by the remote host.
and
> 2023/08/30 03:28:48 [proxy] ERROR: failed to copy inbound CONNECT tunnel: readfrom tcp 192.168.48.243:3128->192.168.48.220:49453: read tcp 192.168.48.243:46234->52.113.205.120:443: wsarecv: An existing connection was forcibly closed by the remote host.

Firewall checked to my best knowledge, all traffic allowed. Sometimes it works for a while, than starts to rain errors. It's safe to assume it's getting worse with time, but not completely gone by a restart.

Any suggestions? Thanks!

bvarnai:
Update: I tried the docker image on Windows, no sign of this error. This works for me, but maybe worth to investigate. At least it proves that it's not caused by local network/firewall or VPN issue.

Choraden:
Hello, thank you for reporting. 
We will investigate this. 

In the meantime, you can try [v1.0.3](https://github.com/saucelabs/forwarder/releases/tag/v1.0.3) that has just been released.

mmatczuk:
Hi @bvarnai the issue is fixed. Could you do me a favor and check if version v1.1-rc1 solves it for you?

bvarnai:
I still experience these errors, however they are no longer flooding the console. Not a single one reported in docker.

mmatczuk:
This error means that the client (browser) has disconnected

```
WSAECONNRESET
10054 (0x2746)
An existing connection was forcibly closed by the remote host.
```
https://learn.microsoft.com/en-us/windows/win32/debug/system-error-codes--9000-11999-

How do you experience the error now? Can you share more information and debug logs.

mmatczuk:
It looks like we need to handle that also in tunnel.
[...]

PR #471, opened 2023-10-18
Fix Windows EOF in tunnel treated as error  

mmatczuk:
Relates to #376


Choraden (approved):
Alternative solution not to nest ifs: 
```go
if errors.As(err, &se); se.Syscall == "wsarecv" || se.Syscall == "wsasend"
``` 
becomes 
```go
if errors.As(err, &se) && (se.Syscall == "wsarecv" || se.Syscall == "wsasend")
```

mmatczuk:
I think the suggested solution can still panic if compiler decides to use the `se` value from before the if block.
This is kind of strange and it did not happen in my tests but it happens in CI.

PR #480, opened 2023-10-25
utils/cobrautil: describe flags unredacted

mmatczuk:
We add Unredacted option to FlagsDescriber to enable getting the raw values when needed.

This is to support https://github.com/mmatczuk/anyflag/pull/2


Choraden (commented):
LGMT, however I'd double check whether we need this. What is the use case of describing unredacted values? Possible security risk.

mmatczuk:
We do.

PR #483, opened 2023-10-26
httplog: Merge the log-http and api-log-http CLI options

alexh-sauce:
Instead of having separate log-http and api-log-http options, merge them into `--log-http [<module>:]<mode>`

This adds a Config type to `httplog` which contains a map of the module name (used a prefix to the log mode in the CLI) to the log mode:

```go
type Config struct {
        ModeOptions []map[string]Mode                                 
        defaults    map[string]Mode                                   
}                                                           
```

Examples from the forwarder commands showing the various options of single module, multiple modules, or no modules.

```
$ ./dist/forwarder_linux_amd64_v1/forwarder run --help 2>&1 | grep "log-http "
    --log-http [<api>|<proxy>:]<none|short-url|url|headers|body|errors> (env FORWARDER_LOG_HTTP)
$ ./dist/forwarder_linux_amd64_v1/forwarder httpbin --help 2>&1 | grep "log-http "
    --log-http [<api>:]<none|short-url|url|headers|body|errors> (env FORWARDER_LOG_HTTP)
$ ± ./dist/forwarder_linux_amd65_v1/forwarder pac server --help 2>&1 | grep "log-http "
    --log-http <none|short-url|url|headers|body|errors> (env FORWARDER_LOG_HTTP)

```

### DIFF
--- a/bind/flag.go
+++ b/bind/flag.go
@@ -7,6 +7,7 @@
 package bind
 
 import (
+	"fmt"
 	"net/netip"
 	"net/url"
 	"os"
@@ -22,6 +23,7 @@ import (
 	"github.com/saucelabs/forwarder/utils/osdns"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+	"golang.org/x/exp/slices"
 )
 
 func ConfigFile(fs *pflag.FlagSet, configFile *string) {
@@ -279,21 +281,47 @@ func HTTPServerConfig(fs *pflag.FlagSet, cfg *forwarder.HTTPServerConfig, prefix
 	fs.VarP(anyflag.NewValueWithRedact[*url.Userinfo](cfg.BasicAuth, &cfg.BasicAuth, forwarder.ParseUserinfo, RedactUserinfo),
 		namePrefix+"basic-auth", "", "<username[:password]>"+
 			"Basic authentication credentials to protect the server. ")
+}
 
-	httpLogModes := []httplog.Mode{
-		httplog.None,
-		httplog.ShortURL,
-		httplog.URL,
-		httplog.Headers,
-		httplog.Body,
-		httplog.Errors,
+func HTTPLogConfig(fs *pflag.FlagSet, cfg []NamedParam[httplog.Mode]) {
+	for _, p := range cfg {
+		if p.Param == nil {
+			panic(fmt.Sprintf("httplog mode is nil for %s", p.Name))
+		}
 	}
-	fs.Var(anyflag.NewValue[httplog.Mode](cfg.LogHTTPMode, &cfg.LogHTTPMode, anyflag.EnumParser[httplog.Mode](httpLogModes...)),
-		namePrefix+"log-http", "<none|short-url|url|headers|body|errors>"+
-			"HTTP request and response logging mode. "+
-			"Setting this to none disables logging. "+
-			"The short-url mode logs [scheme://]host[/path] instead of the full URL. "+
-			"The error mode logs request line and headers if status code is greater than or equal to 500. ")
+
+	names := httplogExtractNames(cfg)
+
+	parse := func(val string) (NamedParam[httplog.Mode], error) {
+		name, mode, err := httplog.SplitNameMode(val)
+		if err != nil {
+			return NamedParam[httplog.Mode]{}, err
+		}
+		if name != "" && !slices.Contains(names, name) {
+			return NamedParam[httplog.Mode]{}, fmt.Errorf("unknown name: %s", name)
+		}
+
+		return NamedParam[httplog.Mode]{Name: name, Param: &mode}, nil
+	}
+
+	var flagValue []NamedParam[httplog.Mode]
+	f := httplogFlag{
+		SliceValue: anyflag.NewSliceValue[NamedParam[httplog.Mode]](flagValue, &flagValue, parse),
+		update: func() {
+			httplogUpdate(cfg, flagValue)
+		},
+	}
+
+	valueType := "<none|short-url|url|headers|body|errors>"
+	if ss := names; len(ss) > 1 {
+		valueType = "[" + strings.Join(ss, "|") + ":]" + valueType
+	}
+
+	fs.Var(f, "log-http", valueType+",... "+
+		"HTTP request and response logging mode. "+
+		"Setting this to none disables logging. "+
+		"The short-url mode logs [scheme://]host[/path] instead of the full URL. "+
+		"The error mode logs request line and headers if status code is greater than or equal to 500. ")
 }
 
 func TLSServerConfig(fs *pflag.FlagSet, cfg *forwarder.TLSServerConfig, namePrefix string) {
--- a/bind/log.go
+++ b/bind/log.go
@@ -51,7 +51,7 @@ func httplogUpdate(dst, src []NamedParam[httplog.Mode]) {
 	}
 
 	// Find default mode if set.
-	defaultMode := httplog.Errors
+	defaultMode := httplog.DefaultMode
 	for i := range src {
 		j := len(src) - i - 1
 		if src[j].Name == "" {
--- a/bind/log.go
+++ b/bind/log.go
@@ -50,8 +50,8 @@ func httplogUpdate(dst, src []NamedParam[httplog.Mode]) {
 		}
 	}
 
-	// Find default mode if set.
-	defaultMode := httplog.DefaultMode
+	// Find default mode.
+	var defaultMode httplog.Mode
 	for i := range src {
 		j := len(src) - i - 1
 		if src[j].Name == "" {
@@ -60,10 +60,12 @@ func httplogUpdate(dst, src []NamedParam[httplog.Mode]) {
 		}
 	}
 
-	// Set default mode for unset values.
-	for i := range dst {
-		if !changed[i] {
-			*dst[i].Param = defaultMode
+	// If default mode is set, update dst with it.
+	if defaultMode != "" {
+		for i := range dst {
+			if !changed[i] {
+				*dst[i].Param = defaultMode
+			}
 		}
 	}
 }
--- a/bind/log.go
+++ b/bind/log.go
@@ -1,0 +1,79 @@
+// Copyright 2023 Sauce Labs Inc. All rights reserved.
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+package bind
+
+import (
+	"github.com/mmatczuk/anyflag"
+	"github.com/saucelabs/forwarder/httplog"
+)
+
+type httplogFlag struct {
+	*anyflag.SliceValue[NamedParam[httplog.Mode]]
+	update func()
+}
+
+func (f httplogFlag) Set(val string) (err error) {
+	err = f.SliceValue.Set(val)
+
+	if err == nil {
+		f.update()
+	}
+
+	return
+}
+
+func (f httplogFlag) Replace(vals []string) (err error) {
+	err = f.SliceValue.Replace(vals)
+
+	if err == nil {
+		f.update()
+	}
+
+	return
+}
+
+func httplogUpdate(dst, src []NamedParam[httplog.Mode]) {
+	changed := make([]bool, len(dst))
+
+	// Update dst with src values.
+	for i := range dst {
+		for j := range src {
+			if dst[i].Name == src[j].Name {
+				*dst[i].Param = *src[j].Param
+				changed[i] = true
+				break
+			}
+		}
+	}
+
+	// Find default mode if set.
+	defaultMode := httplog.Errors
+	for i := range src {
+		j := len(src) - i - 1
+		if src[j].Name == "" {
+			defaultMode = *src[j].Param
+			break
+		}
+	}
+
+	// Set default mode for unset values.
+	for i := range dst {
+		if !changed[i] {
+			*dst[i].Param = defaultMode
+		}
+	}
+}
+
+func httplogExtractNames(cfg []NamedParam[httplog.Mode]) []string {
+	names := make([]string, 0, len(cfg))
+	for _, c := range cfg {
+		if c.Name != "" {
+			names = append(names, c.Name)
+		}
+	}
+	return names
+}
--- a/httplog/httplog.go
+++ b/httplog/httplog.go
@@ -63,6 +63,8 @@ func SplitNameMode(val string) (name string, mode Mode, err error) {
 	return
 }
 
+var DefaultMode = Errors
+
 type Logger struct {
 	log  func(format string, args ...any)
 	mode Mode
@@ -70,6 +72,9 @@ type Logger struct {
 
 // NewLogger returns a logger that logs HTTP requests and responses.
 func NewLogger(logFunc func(format string, args ...any), mode Mode) *Logger {
+	if mode == "" {
+		mode = DefaultMode
+	}
 	return &Logger{
 		log:  logFunc,
 		mode: mode,
--- a/httplog/httplog.go
+++ b/httplog/httplog.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"strings"
 
 	"github.com/saucelabs/forwarder/internal/martian/log"
 	"github.com/saucelabs/forwarder/internal/martian/messageview"
@@ -30,6 +31,36 @@ const (
 
 func (m Mode) String() string {
 	return string(m)
+}
+
+func SplitNameMode(val string) (name string, mode Mode, err error) {
+	n, m, ok := strings.Cut(val, ":")
+	if ok {
+		name = n
+		mode = Mode(m)
+	} else {
+		name = ""
+		mode = Mode(val)
+	}
+
+	switch mode {
+	case None:
+		mode = None
+	case ShortURL:
+		mode = ShortURL
+	case URL:
+		mode = URL
+	case Headers:
+		mode = Headers
+	case Body:
+		mode = Body
+	case Errors:
+		mode = Errors
+	default:
+		return "", "", fmt.Errorf("invalid mode %q", mode)
+	}
+
+	return
 }
 
 type Logger struct {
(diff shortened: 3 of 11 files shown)


Choraden:
This approach kills the usage of default values from servers' configs. 

There is: 
```go
httpProxyConfig:     forwarder.DefaultHTTPProxyConfig(),
...
apiServerConfig:     forwarder.DefaultHTTPServerConfig(),
logHTTPConfig:       httplog.NewConfig(map[string]httplog.Mode{"api": httplog.Errors, "proxy": httplog.Errors})
```
While we already have default `httplog.Mode` specified in `forwarder.DefaultHTTPServerConfig()` and `forwarder.DefaultHTTPProxyConfig()`.

httplog/httplog.go
```diff
@@ -37,6 +39,66 @@ type Logger struct {
 	mode Mode
 }
 
+type Config struct {
+	ModeOptions []map[string]Mode
+	defaults    map[string]Mode
+}
```
Choraden, line 45: I don't like the structure of it. 

Notice that in all other places we don't have `defaults` field. We simply set the values, thus I'd propose something like that

```go
type ModuleMode struct {
  module string
  mode Mode
}

type Config []ModuleMode
```
Default value is the one you set at the beginning of the command.

Choraden: I'm also unsure whether this `Config` should live here. It only collects `http log.Mode`. Maybe `config.go` (forwarder root dir) would be better place.

Choraden:
I have mixed feelings whether we need this. 
On one hand, merging has some advantages. On the other hand, it is only useful when one wants to set all modules to the same mode (which I doubt would be a frequent case), otherwise it would complicate the ease of usability, as user would need to know available modules.

httplog/httplog.go
mmatczuk: Since it needs to parse the log modes it should be where the modes are defined IMO.

httplog/httplog.go
```diff
@@ -37,6 +39,66 @@ type Logger struct {
...
+	if strings.Contains(s, ":") {
+		m := strings.SplitN(s, ":", 2)
+		module = m[0]
+		modeStr = m[1]
+	} else {
+		modeStr = s
+	}
+
+	switch modeStr {
+	case "none":
```
mmatczuk, line 61: Use const value.

httplog/httplog.go
```diff
@@ -37,6 +39,66 @@ type Logger struct {
...
+	ModeOptions []map[string]Mode
+	defaults    map[string]Mode
+}
+
+func ParseMode(s string) (map[string]Mode, error) {
+	var modeStr string
+	var mode Mode
+	module := ""
+
+	if strings.Contains(s, ":") {
```
mmatczuk, line 52: Use strings.Cut

httplog/httplog.go
```diff
@@ -37,6 +39,66 @@ type Logger struct {
...
+		mode = Headers
+	case "body":
+		mode = Body
+	case "errors":
+		mode = Errors
+	default:
+		return map[string]Mode{}, fmt.Errorf("unknown HTTP log mode: %s", s)
+	}
+
+	return map[string]Mode{module: mode}, nil
```
mmatczuk, line 77: Returning map containing single element makes little sense.
Parse should return `ModuleMode` as @Choraden suggested.

httplog/httplog.go
```diff
@@ -37,6 +39,66 @@ type Logger struct {
...
+
+// NewConfig returns a new Config that controls logging levels based on their module name.
+func NewConfig(defaults map[string]Mode) *Config {
+	return &Config{
+		ModeOptions: []map[string]Mode{},
+		defaults:    defaults,
+	}
+}
+
+func (m *Config) GetModules() []string {
```
mmatczuk, line 88: We do not use GetX but X instead.

mmatczuk: Do we need it?

httplog/httplog.go
```diff
@@ -37,6 +39,66 @@ type Logger struct {
...
+	return maps.Keys(m.defaults)
+}
+
+func (m *Config) GetMode(module string) Mode {
```
mmatczuk, line 92: We are missing fallback rule, looking for module M
- if M:<mode> specified used that
- if global <mode> specified use that
- otherwise return Errors

This should have a test.

httplog/httplog.go
alexh-sauce: Ok, thanks for the suggestion. I implemented it as you described but realized we also need to check for valid module names when parsing, so I also added a slice to track valid module names. WDYT?

```go
type Config struct {
       ModuleModes []ModuleMode
       modules     []string
}
```

alexh-sauce:
> This approach kills the usage of default values from servers' configs.
> 
> While we already have default `httplog.Mode` specified in `forwarder.DefaultHTTPServerConfig()` and `forwarder.DefaultHTTPProxyConfig()`.

Ah, that was an oversight. Should be fixed now by skipping the call to set it if not found.

bind/flag.go
```diff
@@ -279,17 +279,25 @@ func HTTPServerConfig(fs *pflag.FlagSet, cfg *forwarder.HTTPServerConfig, prefix
...
+	valid := make([]string, 0, len(logcfg.Modules()))
+	for _, m := range logcfg.Modules() {
+		if m != "" {
+			valid = append(valid, m)
+		}
 	}
-	fs.Var(anyflag.NewValue[httplog.Mode](cfg.LogHTTPMode, &cfg.LogHTTPMode, anyflag.EnumParser[httplog.Mode](httpLogModes...)),
-		namePrefix+"log-http", "<none|short-url|url|headers|body|errors>"+
+
+	modulesStr := ""
```
mmatczuk, line 292: Use `strings.Builder`

bind/flag.go
```diff
@@ -279,17 +279,25 @@ func HTTPServerConfig(fs *pflag.FlagSet, cfg *forwarder.HTTPServerConfig, prefix
...
 
-	httpLogModes := []httplog.Mode{
-		httplog.None,
-		httplog.ShortURL,
-		httplog.URL,
-		httplog.Headers,
-		httplog.Body,
-		httplog.Errors,
+func HTTPLogConfig(fs *pflag.FlagSet, logcfg *httplog.Config) {
+	valid := make([]string, 0, len(logcfg.Modules()))
```
mmatczuk, line 285: I do not see why the config should know the modules why not pass the modules directly as a parameter here?

httplog/httplog.go
```diff
@@ -37,6 +39,71 @@ type Logger struct {
 	mode Mode
 }
 
+type ModuleMode struct {
+	Module string
+	mode   Mode
```
mmatczuk, line 44: This can be exported.

httplog/httplog.go
```diff
@@ -37,6 +39,71 @@ type Logger struct {
...
+}
+
+type Config struct {
```
mmatczuk, line 47: I'd say config can be `type Config  []ModuleMode`

bind/flag.go
mmatczuk: I get it now, this way it can validate.

bind/flag.go
```diff
@@ -279,17 +279,25 @@ func HTTPServerConfig(fs *pflag.FlagSet, cfg *forwarder.HTTPServerConfig, prefix
...
+	for _, m := range logcfg.Modules() {
```
mmatczuk, line 286: There is no need to copy it here.
If you want a copy for safety, just copy the slice before returning it in Modules.

httplog/httplog.go
```diff
@@ -37,6 +39,71 @@ type Logger struct {
...
+	ModuleModes []ModuleMode
+	modules     []string
+}
+
+func (m *Config) ParseMode(s string) (ModuleMode, error) {
+	var mode Mode
+
+	module, modeStr, found := strings.Cut(s, ":")
```
mmatczuk, line 55: Standard naming is  `ok`.

httplog/httplog.go
```diff
@@ -37,6 +39,71 @@ type Logger struct {
...
+
```
mmatczuk, line 56: Standard layout gives no light here.

httplog/httplog.go
```diff
@@ -37,6 +39,71 @@ type Logger struct {
...
+	if !found {
+		module = ""
+		modeStr = s
+	}
+
+	// First, check if the module is valid
+	if !slices.Contains(m.modules, module) {
```
mmatczuk, line 63: Good!

httplog/httplog.go
```diff
@@ -37,6 +39,71 @@ type Logger struct {
...
+// NewConfig returns a new Config that controls logging levels based on their module name.
+func NewConfig(modules []string) *Config {
+	return &Config{modules: modules}
+}
+
+func (m *Config) Modules() []string {
+	return m.modules
+}
+
+func (m *Config) Mode(module string) (Mode, error) {
```
mmatczuk, line 97: You need to check for global mode too.

httplog/httplog_test.go
```diff
@@ -0,0 +1,90 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+package httplog
+
+import (
+	"testing"
+)
+
+func TestParseMode(t *testing.T) {
```
mmatczuk, line 13: Often it's easier to have two tests one for OK situations other for error validation.

mmatczuk, line 13: Should be TestConfigXXX

mmatczuk:
Regarding linting please rebase

Choraden:
In order to "register" a module we need to manually set it here 
```go
httplog.NewConfig([]string{"", "api"})
```
and later use it also manually
```go
if mode, err := c.logHTTPConfig.Mode("api"); err == nil {
 	c.apiServerConfig.LogHTTPMode = mode
 }
 if mode, err := c.logHTTPConfig.Mode(""); err == nil {
 	c.httpServerConfig.LogHTTPMode = mode
 }
 ```
 I was hoping we could have a solution that would be more flexible e.g. use a pointer to the mode to modify the modes directly when assigning values in flags. 
 ```go
 ModuleMode{
 Module: "api",
 Mode: &c.apiServerConfig.LogHTTPMode,
}
 ```
This would remove the step where we manually assign modes from the config. This could also remove "allowed modules" field as we would have all available modules in the initial slice. *Details need to be polished.

File: bind/flag.go (hunk 11 above)
Comment (alexh-sauce) [reply bind/flag.go:286]: It's not just a copy though, it's removing the empty string.

If we don't do that we get a usage string like `[<>|<api>:]` or `[<>]`. I believe the latter case confuses Cobra and it doesn't print proper usage text.

Or do you mean do the filtering inside the `Modules()` function?

File: httplog/httplog.go (hunk 12 above)
Comment (alexh-sauce) [reply httplog/httplog.go:55]: Sorry, I don't understand this comment - the function signature is 

```go
func Cut(s, sep string) (before, after string, found bool)
```

?

alexh-sauce:
@Choraden 

> I was hoping we could have a solution that would be more flexible e.g. use a pointer to the mode to modify the modes directly when assigning values in flags.

I agree that would be nice, but I don't see a way we could do that *and* get a global/default fallback. From what I can see, the mode wouldn't be calculated if it was missing, so there wouldn't be anywhere to trigger the global/default fallback. Happy to be corrected if I'm wrong here.

httplog/httplog.go
mmatczuk: Interesting, usually the when bool is returned it's called `ok`

httplog/httplog.go
```diff
@@ -37,6 +39,82 @@ type Logger struct {
...
+func NewConfig(modules []string) *Config {
+	return &Config{modules: modules}
+}
+
+func (c *Config) Modules() []string {
+	return c.modules
+}
+
+func (c *Config) Mode(module string) Mode {
+	// First, check if the module matches exactly
```
mmatczuk, line 98: Can we finish with a dot.

mmatczuk:
I like the idea, let's ignore fallback to "" for now and make the module mandatory.

On that front can we have Name instead of Module?
Module seems like go module which is a bit misleading.

mmatczuk:
Regarding CI it looks like the local golangci-lint version is compiled with go 1.21 while the CI version is compiled with 1.20. I bumped the version. Please rebase again @alexh-sauce.

Choraden:
I don't think there should be a problem with fallback, but maybe I don't fully see it. 

Let me share my idea in a pseudo code:

```go
type Config []ModuleMode

// Implement pflag.Value interface on Config.

func (c Config) String() string {
	return "implement me"
}

func (c Config) Set(s string) error {
	n, m, ok := strings.Cut(s, ":")
	if !ok {
		n = ""
		m = s
	}

	for i := range c {
		if c[i].Module == n {
			*c[i].Mode = Mode(m)
			return nil
		}
	}

	return fmt.Errorf("module %q not found", n)
}

func (c Config) Type() string {
	return "httplog.Config"
}
```
Config would consist of modules and their default modes. The only thing user can do is to modify underlying mode value of some module by calling `Set`.

alexh-sauce:
Ok, implemented the `SliceValue` interface, and as a bonus, the default values show up in the usage text too. Global defaults are unused, we either use the value set in the flag for the specific module, or we use the default that the `Mode` it set to in the setup of the `Config` instance.

```go
c.logHTTPConfig = &httplog.Config{
		{Name: "api", Mode: &c.apiServerConfig.LogHTTPMode},
		{Name: "", Mode: &c.httpServerConfig.LogHTTPMode},
	}
```

httplog/httplog.go
```diff
@@ -79,40 +61,83 @@ func (c *Config) ParseMode(s string) (ModuleMode, error) {
...
+func (c *Config) Replace(s []string) error {
+	c = nil
+
+	for _, m := range s {
+		n, err := NewModuleMode(m)
+		if err != nil {
+			return err
 		}
+		*c = append(*c, n)
 	}
```
Choraden, line 99: This will panic as you deref `c` which was assigned with nil.

Choraden:
I run it and it's looking good. 

Please squash the commits.

Let's change the flag description to help users understand what's going on with modules :).

bind/flag.go
```diff
@@ -279,17 +279,25 @@ func HTTPServerConfig(fs *pflag.FlagSet, cfg *forwarder.HTTPServerConfig, prefix
...
+			valid = append(valid, m.Name)
+		}
 	}
-	fs.Var(anyflag.NewValue[httplog.Mode](cfg.LogHTTPMode, &cfg.LogHTTPMode, anyflag.EnumParser[httplog.Mode](httpLogModes...)),
-		namePrefix+"log-http", "<none|short-url|url|headers|body|errors>"+
+
+	var b strings.Builder
+	if len(valid) > 0 {
+		b.WriteString("[<")
+		b.WriteString(strings.Join(valid, ">|<"))
```
mmatczuk, line 295: You do not need `strings.Join`

bind/flag.go
```diff
@@ -279,17 +279,25 @@ func HTTPServerConfig(fs *pflag.FlagSet, cfg *forwarder.HTTPServerConfig, prefix
...
-		httplog.Headers,
-		httplog.Body,
-		httplog.Errors,
+func HTTPLogConfig(fs *pflag.FlagSet, logcfg *httplog.Config) {
+	valid := make([]string, 0, len(*logcfg))
+	for _, m := range *logcfg {
+		if m.Name != "" {
+			valid = append(valid, m.Name)
+		}
 	}
```
mmatczuk, line 290: The string building should happen in this for loop. We do not need to rewrite the slice.

httplog/httplog.go
```diff
@@ -32,6 +33,108 @@ func (m Mode) String() string {
...
+	if err != nil {
+		return err
+	}
+
+	for i := range c {
+		if c[i].Name == m.Name {
+			*c[i].Mode = *m.Mode
+			return nil
+		}
+	}
```
mmatczuk, line 129: If there is no fallback to "" we can have anyflag implementation and avoid the boilerplate.

I'm thinking that you could also have a wrapper type around anyflag type that would overwrite Set and Replace adding post processing i.e. if we have default specified and the flag is unset use the default.

httplog/httplog.go
```diff
@@ -32,6 +33,108 @@ func (m Mode) String() string {
 	return string(m)
 }
 
+type ModuleMode struct {
+	Name string
+	Mode *Mode
+}
+
+func NewModuleMode(s string) (ModuleMode, error) {
```
mmatczuk, line 41: This is usually named by Parse

mmatczuk:
I added a commit that demonstrates how to wrap any type.

I think that httplog.Config can be removed - it serves no purpose

mmatczuk:
I requires some more work @Choraden.

bind/log.go
```diff
@@ -0,0 +1,79 @@
...
+			if dst[i].Name == src[j].Name {
+				*dst[i].Param = *src[j].Param
+				changed[i] = true
+				break
+			}
+		}
+	}
+
+	// Find default mode if set.
+	defaultMode := httplog.Errors
```
Choraden, line 54: How about not setting default mode if it hasn't been found. 
Right now we ignore default modes from config.

mmatczuk: It's less effective. Now you can set headers and you would get headers everywhere, you can do `headers,api:none` all but api etc.

Choraden: But if you do `--log-http api:body` it would change all the other modes to `httplog.Errors` which is a default in update function. This ignores default values for other modes set there:
```go
bind.HTTPLogConfig(fs, []bind.NamedParam[httplog.Mode]{
		{Name: "api", Param: &c.apiServerConfig.LogHTTPMode},
		{Name: "proxy", Param: &c.httpProxyConfig.LogHTTPMode},
	})
```
Am I missing something?

mmatczuk: I guess the missing part is that it's irrelevant because the default is Errors.
We could formalize that, remove the default values from DefaultConfigs and make "" resolve to Errors.

mmatczuk: Or add DefaultMode.

Choraden: It's irrelevant in this case, but it might be a trouble in sc, where we have a different default mode than Errors. 

Allowing user to choose a DefaultMode would be decent.

mmatczuk: Check the followup comments, now you can set the global default, and the defaulting in bind was removed.

Choraden (approved):

PR #484, opened 2023-10-30
Auto GOMAXPROCS

mmatczuk:


cmd/forwarder/run/run.go
```diff
@@ -67,6 +69,9 @@ func (c *command) runE(cmd *cobra.Command, _ []string) (cmdErr error) {
 
 	var ep []forwarder.APIEndpoint
 
+	logger.Infof("Forwarder %s (%s)", version.Version, version.Commit)
+	logger.Debugf("resource limits: GOMAXPROCS=%d GOMEMLIMIT=%s", runtime.GOMAXPROCS(0), os.Getenv("GOMEMLIMIT"))
```
Choraden, line 73: nit: I don't like this empty log `GOMEMLIMIT=`, even if it's just a debug mode.

Choraden:
LGTM

cmd/forwarder/run/run.go
mmatczuk: Would you like it like `<unknown>`

Choraden: Yes, please.

mmatczuk:
I changed it so that it only prints GOMEMLIMIT when set.

PR #490, opened 2023-10-31
Improve packaging and rework local dev

Choraden:


mmatczuk:
You added `packaging/forwarder.env` in packaging: remove NOFILE limit

Choraden:
I added only a commented out `GOMEMLIMIT` example to balance the removal of NOFILE limit. Shouldn't it be there?

mmatczuk:
It would be better to add that in two commits.

mmatczuk:
Thanks.

PR #492, opened 2023-11-02
utils/cobrautil: make describe return []bytes insted of string

### LINKED ISSUE #481: Change cobrautil.DescribeFlags to retun []byte instead of string
Usually what we want is bytes. Converting back and forth requires rewriting it. It makes no sense.

Choraden:
Usually what we want is bytes.


mmatczuk:
LGTM

PR #496, opened 2023-11-02
Improve domain access denied error

Choraden:


http_proxy.go
```diff
@@ -445,23 +445,27 @@ func (hp *HTTPProxy) basicAuth(u *url.Userinfo) martian.RequestModifier {
...
 func (hp *HTTPProxy) denyDomains(r *ruleset.RegexpMatcher) martian.RequestModifier {
-	return hp.abortIf(func(req *http.Request) bool {
-		return r.Match(req.URL.Hostname())
-	}, func(req *http.Request) *http.Response {
-		return hp.errorResponse(req, ErrProxyDenied)
-	}, errors.New("domain access denied"))
+	return hp.abortIf(
+		func(req *http.Request) bool { return r.Match(req.URL.Hostname()) },
+		func(req *http.Request) *http.Response { return hp.errorResponse(req, ErrProxyDenied) },
+		func(req *http.Request) error { return fmt.Errorf("access to %v denied", req.URL.Hostname()) },
```
mmatczuk, line 467: Why %v not %s?

Choraden:
I'm afraid, changing log level to info won't be that easy. The error comes from request modifier and is logged there - that's how this abort mechanism works. Good that you extracted it to #498. Let's send another patch then.

PR #502, opened 2023-11-03
http_proxy: print proxy errors in response body

Choraden:
Responses will now contain proxy name and error:
First line is a generic message, second the error.
```
curl -x localhost:3128 not.exist   
forwarder failed to connect to remote host
dial tcp: lookup not.exist on 192.168.1.1:53: no such host
```

Header:
```
X-Forwarder-Error: forwarder dial tcp: lookup www on 10.255.3.12:53: no such host
```

### DIFF
--- a/e2e/tests/proxy_test.go
+++ b/e2e/tests/proxy_test.go
@@ -235,9 +235,30 @@ func TestProxyBadGateway(t *testing.T) {
 		"httpbin:1",
 	}
 
+	var expectedErrorMessage string
+	switch {
+	case os.Getenv("FORWARDER_PROXY") != "":
+		expectedErrorMessage = forwarder.UpstreamProxyServiceName + " failed to connect to remote host"
+	case os.Getenv("FORWARDER_PAC") != "":
+		// Proxy name depends on the PAC file. Skip it for now.
+		expectedErrorMessage = "failed to connect to remote host"
+	default:
+		expectedErrorMessage = "forwarder failed to connect to remote host"
+	}
+
 	for _, scheme := range []string{"http", "https"} {
 		for _, h := range hosts {
-			newClient(t, scheme+"://"+h).GET("/status/200").ExpectStatus(http.StatusBadGateway)
+			t.Run(scheme+"_"+h, func(t *testing.T) {
+				res := newClient(t, scheme+"://"+h).GET("/status/200")
+				res.ExpectStatus(http.StatusBadGateway)
+
+				// Check if the error message is correctly propagated to the client.
+				// Especially when several proxies are chained.
+				// FIXME(hg): When HTTPS CONNECT request fails it does not propagate error message - HTTP client does not return it.
+				if scheme == "http" && !strings.Contains(string(res.Body), expectedErrorMessage) {
+					t.Fatalf("Expected valid error message, got %s", res.Body)
+				}
+			})
 		}
 	}
 }
--- a/http_proxy_errors.go
+++ b/http_proxy_errors.go
@@ -50,7 +50,7 @@ func (hp *HTTPProxy) errorResponse(req *http.Request, err error) *http.Response 
 	}
 	if code == 0 {
 		code = http.StatusInternalServerError
-		msg = "An unexpected error occurred"
+		msg = "encountered an unexpected error"
 		label = "unexpected_error"
 	}
 
@@ -78,10 +78,10 @@ func handleNetError(_ *http.Request, err error) (code int, msg, label string) {
 	if errors.As(err, &netErr) {
 		if netErr.Timeout() {
 			code = http.StatusGatewayTimeout
-			msg = "Timed out connecting to remote host"
+			msg = "timed out connecting to remote host"
 		} else {
 			code = http.StatusBadGateway
-			msg = "Failed to connect to remote host"
+			msg = "failed to connect to remote host"
 		}
 		label = "net_" + netErr.Op
 	}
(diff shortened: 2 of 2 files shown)


http_proxy_errors.go
```diff
@@ -56,10 +56,11 @@ func (hp *HTTPProxy) errorResponse(req *http.Request, err error) *http.Response
 
 	hp.metrics.error(label)
 
-	resp := proxyutil.NewResponse(code, bytes.NewBufferString(msg+"\n"), req)
+	body := fmt.Sprintf("%s: %s\n%s\n", hp.config.Name, msg, err)
+	resp := proxyutil.NewResponse(code, bytes.NewBufferString(body), req)
```
mmatczuk, line 60: I see ya learning to use bytes pkg :)

mmatczuk:
Can it be like 

```
forwarder failed to connect to remote host
dial tcp: lookup not.exist on 192.168.1.1:53: no such host
```

mmatczuk:
Also, check the response headers, the proxy name must be both in headers and body.

http_proxy_errors.go
```diff
@@ -56,10 +57,18 @@ func (hp *HTTPProxy) errorResponse(req *http.Request, err error) *http.Response
 
 	hp.metrics.error(label)
 
-	resp := proxyutil.NewResponse(code, bytes.NewBufferString(msg+"\n"), req)
-	resp.Header.Set(ErrorHeader, err.Error())
+	var body bytes.Buffer
+	body.WriteString(hp.config.Name)
+	body.WriteString(" ")
+	body.WriteString(strings.ToLower(msg))
```
mmatczuk, line 63: How about we fixed the messages so that they do not start with upper case?

Choraden: ok

http_proxy_errors.go
```diff
@@ -121,7 +130,7 @@ func handleStatusText(req *http.Request, err error) (code int, msg, label string
 	if req.URL.Scheme == "https" && err != nil {
 		for i := 400; i < 600; i++ {
 			if err.Error() == http.StatusText(i) {
-				return i, err.Error(), "https_status_text"
+				return i, strings.ToLower(err.Error()), "https_status_text"
```
mmatczuk, line 133: let's add some text here, "encountered an unexpected internal error: <error>"

http_proxy_errors.go
```diff
@@ -121,7 +130,8 @@ func handleStatusText(req *http.Request, err error) (code int, msg, label string
 	if req.URL.Scheme == "https" && err != nil {
 		for i := 400; i < 600; i++ {
 			if err.Error() == http.StatusText(i) {
-				return i, err.Error(), "https_status_text"
+				msg = fmt.Sprintf("encountered an unexpected internal error: %s", strings.ToLower(err.Error()))
```
mmatczuk, line 133: So do we need `strings.ToLower` here?

http_proxy_errors.go
```diff
@@ -50,16 +51,24 @@ func (hp *HTTPProxy) errorResponse(req *http.Request, err error) *http.Response
 	}
 	if code == 0 {
 		code = http.StatusInternalServerError
-		msg = "An unexpected error occurred"
+		msg = "an unexpected error occurred"
```
mmatczuk, line 54: This does not play well with config.Name

e2e/tests/proxy_test.go
```diff
@@ -235,9 +235,26 @@ func TestProxyBadGateway(t *testing.T) {
...
 	for _, scheme := range []string{"http", "https"} {
 		for _, h := range hosts {
-			newClient(t, scheme+"://"+h).GET("/status/200").ExpectStatus(http.StatusBadGateway)
+			t.Run(scheme+"_"+h, func(t *testing.T) {
+				res := newClient(t, scheme+"://"+h).GET("/status/200")
+				res.ExpectStatus(http.StatusBadGateway)
+
+				// Check if the error message is correctly propagated to the client.
+				// Especially when several proxies are chained.
+				if !strings.Contains(string(res.Body), expectedErrorMessage) {
```
mmatczuk, line 253: Nah, just skip it if MITM is enabled.

Choraden: The issue is that our current httpexpect implementation also does the same trick, so all https requests made by the client do not return an original response in case of a connect error. 
```go
func (c *Client) do(req *http.Request) (*http.Response, error) {
	if c.trace != nil {
		req = req.WithContext(httptrace.WithClientTrace(req.Context(), c.trace))
	}
	resp, err := c.rt.RoundTrip(req)

	// There is a difference between sending HTTP and HTTPS requests.
	// For HTTPS client issues a CONNECT request to the proxy and then sends the original request.
	// In case the proxy responds with status code 4XX or 5XX to the CONNECT request, the client interprets it as URL error.
	//
	// This is to cover this case.
	if req.URL.Scheme == "https" && err != nil {
		for i := 400; i < 600; i++ {
			if err.Error() == http.StatusText(i) {
				return &http.Response{
					StatusCode: i,
					Status:     http.StatusText(i),
					ProtoMajor: 1,
					ProtoMinor: 1,
					Header:     http.Header{},
					Body:       http.NoBody,
					Request:    req,
				}, nil
			}
		}
	}

	return resp, err
}
```

Solution 1. is to keep this test and later address it in both proxy http client and our http expect client.

Solution 2. is to limit the scope and make it a dedicated MITM flag test.

Choraden: I actually wanted to print a warning there, but I'm afraid, it won't be visible in a non-debug mode. So there we have this "half skip" - it always checks the status code.

mmatczuk: Add text checking only for http requests.

Choraden: ok

PR #513, opened 2023-11-09
e2e: do not log http in api module while in debug mode

Choraden:
API servers receive healthcheck requests which pollutes the logs.


mmatczuk:
I'd suggest headers, api:errors

PR #517, opened 2023-11-09
Genetate default config file

### LINKED ISSUE #504: docs: generate default config file

mmatczuk:
This allows to automatically produce the following config file

```
Created symlink /etc/systemd/system/multi-user.target.wants/forwarder.service → /usr/lib/systemd/system/forwarder.service.
● forwarder.service - Sauce Labs Forwarder Service
     Loaded: loaded (/usr/lib/systemd/system/forwarder.service; enabled; vendor preset: enabled)
     Active: active (running) since Thu 2023-11-09 13:13:41 UTC; 100ms ago
   Main PID: 61 (forwarder)
        CPU: 5ms
     CGroup: /system.slice/forwarder.service
             └─61 /usr/bin/forwarder run

Nov 09 13:13:41 ad796138a0d5 systemd[1]: Started Sauce Labs Forwarder Service.
Nov 09 13:13:41 ad796138a0d5 forwarder[61]: 2023/11/09 13:13:41.938072 [INFO] Forwarder 1.1-rc1-SNAPSHOT-10c27618 (10c27618d1805e970927f045786ceec40c69de73)
Nov 09 13:13:41 ad796138a0d5 forwarder[61]: 2023/11/09 13:13:41.938258 [INFO] configuration
Nov 09 13:13:41 ad796138a0d5 forwarder[61]: config-file=/etc/forwarder/forwarder.yaml
Nov 09 13:13:41 ad796138a0d5 forwarder[61]: 2023/11/09 13:13:41.938345 [proxy] [INFO] no upstream proxy specified
Nov 09 13:13:41 ad796138a0d5 forwarder[61]: 2023/11/09 13:13:41.938364 [proxy] [INFO] localhost proxying mode=deny
Nov 09 13:13:41 ad796138a0d5 forwarder[61]: 2023/11/09 13:13:41.938622 [proxy] [INFO] PROXY server listen address=[::]:3128 protocol=http
Nov 09 13:13:41 ad796138a0d5 forwarder[61]: 2023/11/09 13:13:41.938933 [api] [INFO] HTTP server listen address=127.0.0.1:10000 protocol=http
root@ad796138a0d5:/# cat /etc/forwarder/forwarder.yaml 
# The server address to listen on. If the host is empty, the server will listen
# on all available interfaces.
#address: :3128

# Basic authentication credentials to protect the server.
#basic-auth:

# Site or upstream proxy basic authentication credentials. The host and port can
# be set to "*" to match all hosts and ports respectively. The flag can be
# specified multiple times to add multiple credentials.
#credentials:

# Name of this proxy instance. This value is used in the Via header in requests.
# The name value in Via header is extended with a random string to avoid
# collisions when several proxies are chained.
#name: forwarder

# The server protocol. For https and h2 protocols, if TLS certificate is not
# specified, the server will use a self-signed certificate.
#protocol: http

# The amount of time allowed to read request headers.
#read-header-timeout: 1m0s

# Global read rate limit in bytes per second i.e. how many bytes per second you
# can receive from a proxy. Accepts binary format (e.g. 1.5Ki, 1Mi, 3.6Gi).
#read-limit: 0

# TLS certificate to use if the server protocol is https or h2. Can be a path to
# a file or "data:" followed by base64 encoded certificate.
#tls-cert-file:

# TLS private key to use if the server protocol is https or h2. Can be a path to
# a file or "data:" followed by base64 encoded key.
#tls-key-file:

# Global write rate limit in bytes per second i.e. how many bytes per second you
# can send to proxy. Accepts binary format (e.g. 1.5Ki, 1Mi, 3.6Gi).
#write-limit: 0

# Deny requests to the specified domains. Prefix domains with '-' to exclude
# requests to certain domains from being denied.
#deny-domains:

# Connect directly to the specified domains without using the upstream proxy.
# Prefix domains with '-' to exclude requests to certain domains from being
# directed.This flag takes precedence over the PAC script.
#direct-domains:

# Add or remove HTTP request headers. Use the format "name: value" to add a
# header, "name;" to set the header to empty value, "-name" to remove the
# header, "-name*" to remove headers by prefix. The header name will be
# normalized to canonical form. The header value should not contain any newlines
# or carriage returns. The flag can be specified multiple times. Example: -H
# "Host: example.com" -H "-User-Agent" -H "-X-*".
#header:

# Proxy Auto-Configuration file to use for upstream proxy selection. It can be a
# local file or a URL, you can also use '-' to read from stdin. The data URI
# scheme is supported, the format is data:base64,<encoded data>.
#pac:

# Upstream proxy to use. The supported protocols are: http, https, socks5. No
# protocol specified will be treated as HTTP proxy. If the port number is not
# specified, it is assumed to be 1080. The basic authentication username and
# password can be specified in the host string e.g. user:pass@host:port.
# Alternatively, you can use the -c, --credentials flag to specify the
# credentials. If both are specified, the proxy flag takes precedence.
#proxy:

# Add or remove HTTP headers on the CONNECT request to the upstream proxy. See
# the documentation for the -H, --header flag for more details on the format.
#proxy-header:

# Setting this to allow enables sending requests to localhost through the
# upstream proxy. Setting this to direct sends requests to localhost directly
# without using the upstream proxy. By default, requests to localhost are
# denied.
#proxy-localhost: deny

# Add or remove HTTP headers on the received response before sending it to the
# client. See the documentation for the -H, --header flag for more details on
# the format.
#response-header:

# Enable Man-in-the-Middle (MITM) mode. It only works with HTTPS requests,
# HTTP/2 is not supported. MITM is enabled by default when the
# --mitm-cacert-file flag is set. If the CA certificate is not provided MITM
# uses a generated CA certificate. The CA certificate used can be retrieved from
# the API server .
#mitm: false

# CA certificate file to use for generating MITM certificates. If the file is
# not specified, a generated CA certificate will be used. See the documentation
# for the --mitm flag for more details.
#mitm-cacert-file:

# CA key file to use for generating MITM certificates.
#mitm-cakey-file:

# Limit MITM to the specified domains. Prefix domains with '-' to exclude
# requests to certain domains from being MITMed.
#mitm-domains:

# Organization name to use in the generated MITM certificates.
#mitm-org: Sauce Labs Inc.

# Validity period of the generated MITM certificates. 
#mitm-validity: 24h0m0s

# If more than one DNS server is specified with the --dns-server flag, passing
# this flag will enable round-robin selection. 
#dns-round-robin: false

# DNS server(s) to use instead of system default. There are two execution
# policies, when more then one server is specified. Fallback: the first server
# in a list is used as primary, the rest are used as fallbacks. Round robin: the
# servers are used in a round-robin fashion. The port is optional, if not
# specified the default port is 53.
#dns-server:

# Timeout for dialing DNS servers. Only used if DNS servers are specified. 
#dns-timeout: 5s

# Add your own CA certificates to verify against. The system root certificates
# will be used in addition to any certificates in this list. Can be a path to a
# file or "data:" followed by base64 encoded certificate. Use this flag multiple
# times to specify multiple CA certificate files.
#cacert-file:

# The maximum amount of time a dial will wait for a connect to complete. With or
# without a timeout, the operating system may impose its own earlier timeout.
# For instance, TCP timeouts are often around 3 minutes. 
#http-dial-timeout: 10s

# The maximum amount of time an idle (keep-alive) connection will remain idle
# before closing itself. Zero means no limit. 
#http-idle-conn-timeout: 1m30s

# The amount of time to wait for a server's response headers after fully writing
# the request (including its body, if any).This time does not include the time
# to read the response body. Zero means no limit. 
#http-response-header-timeout: 0s

# The maximum amount of time waiting to wait for a TLS handshake. Zero means no
# limit.
#http-tls-handshake-timeout: 10s

# Don't verify the server's certificate chain and host name. Enable to work with
# self-signed certificates. 
#insecure: false

# The server address to listen on. If the host is empty, the server will listen
# on all available interfaces.
#api-address: localhost:10000

# Basic authentication credentials to protect the server.
#api-basic-auth:

# The amount of time allowed to read request headers.
#api-read-header-timeout: 1m0s

# Prometheus namespace to use for metrics. The metrics are available at /metrics
# endpoint in the API server.
#prom-namespace: forwarder

# Path to the log file, if empty, logs to stdout.
#log-file:

# HTTP request and response logging mode. Setting this to none disables logging.
# The short-url mode logs [scheme://]host[/path] instead of the full URL. The
# error mode logs request line and headers if status code is greater than or
# equal to 500.
#log-http:

# If the header is present in the request, the proxy will associate the value
# with the request in the logs.
#log-http-request-id-header: X-Request-Id

# Log level.
#log-level: info

root@ad796138a0d5:/#       
```

Fixes #504


utils/cobrautil/templates/yaml_flag_printer.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+}
+
+func (p *YamlFlagPrinter) PrintHelpFlag(f *pflag.Flag) {
+	formatBuf := new(bytes.Buffer)
+	writeYamlFlag(formatBuf, f)
+
+	wrappedStr := formatBuf.String()
+	flagAndUsage := strings.Split(formatBuf.String(), "\n")
+	flagStr := flagAndUsage[len(flagAndUsage)-1]
```
Choraden, line 37: `flagStr` could be moved inside if - it's not used anywhere else.

utils/cobrautil/templates/yaml_flag_printer.go
```diff
@@ -0,0 +1,69 @@
...
+		nextLines := strings.Join(flagAndUsage[:len(flagAndUsage)-1], " ")
+		wrappedUsages := wordwrap.WrapString(nextLines, p.wrapLimit-2)
+		wrappedUsages = "# " + strings.ReplaceAll(wrappedUsages, "\n", "\n# ")
+		wrappedStr = wrappedUsages + "\n#" + flagStr
+	}
+	fmt.Fprintf(p.out, wrappedStr)
+	fmt.Fprintf(p.out, "\n\n")
+}
+
+// writeFlag will output the help flag based
```
Choraden, line 50: outdated comment - function name is different

utils/cobrautil/config_file.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/saucelabs/forwarder/utils/cobrautil/templates"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+)
+
+func ConfigFileCommand(g templates.FlagGroups, fs *pflag.FlagSet) *cobra.Command {
+	return &cobra.Command{
+		Use:    "config-file",
+		Hidden: true,
+		Run: func(cmd *cobra.Command, args []string) {
```
Choraden, line 19: nit: args is not used -> change to `_`

Choraden:
LGTM. It takes time to grasp printer code, but I guess, it's the "trait" of templater.

Left a couple of small comments.

utils/cobrautil/templates/yaml_flag_printer.go
mmatczuk: Inlined.

utils/cobrautil/templates/yaml_flag_printer.go
mmatczuk: Removed.

utils/cobrautil/config_file.go
mmatczuk: OK. Also added `Args:   cobra.NoArgs,`

mmatczuk:
@Choraden PTAL

Choraden (approved):
Inlined looks better 👍

PR #519, opened 2023-11-10
Add prometheus namespace validation 

Choraden:
```
go run ./cmd/forwarder run --prom-namespace upstream-proxy
Error: invalid argument "upstream-proxy" for "--prom-namespace" flag: invalid namespace: upstream-proxy, it must match "^[a-zA-Z_:][a-zA-Z0-9_:]*$"
See 'forwarder run --help' for usage.
exit status 1
```

### DIFF
--- a/config.go
+++ b/config.go
@@ -13,6 +13,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strconv"
 	"strings"
 	_ "unsafe" // for go:linkname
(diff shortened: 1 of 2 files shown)


mmatczuk:
It feels like [http_server: validate prometheus namespace](https://github.com/saucelabs/forwarder/pull/519/commits/3e768eab7c8921df705c3259800408622057a487) goes beyond scope of http server validation.

I'd prefer to limit that to binding.

config.go
```diff
@@ -258,3 +259,19 @@ func OpenFileParser(flag int, perm, dirPerm os.FileMode) func(val string) (*os.F
...
 	}
 }
+
+func validatePrometheusNamespace(val string) error {
+	if val == "" {
+		return nil
+	}
+
+	// https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels
+	pattern := "^[a-zA-Z_:][a-zA-Z0-9_:]*$"
```
mmatczuk, line 277: As a global var we'd benefit from comp time validation.

Choraden:
I limited the verification to binding only.

PR #522, opened 2023-11-14
Instrument error logs

mmatczuk:


cmd/forwarder/run/run.go
```diff
@@ -256,6 +260,22 @@ func (c *command) runE(cmd *cobra.Command, _ []string) (cmdErr error) {
...
+		Namespace: c.httpProxyConfig.PromNamespace,
+		Name:      "errors_total",
+		Help:      "Number of errors",
+	}, []string{"name"})
+
+	if err := c.promReg.Register(m); err != nil {
+		return nil, err
+	}
+
+	return func(name, format string) {
```
Choraden, line 274: nit: `format` is not used -> `_`

Choraden: Don't we have a linter for that? 🤔

Choraden (approved):
LGTM

nit: I would squash `log/stdlog: add OnError()` with `log/stdlog: add name and format to onError`, I don't see a value of keeping them split.

mmatczuk:
Yeah I should rewrite it.

cmd/forwarder/run/run.go
mmatczuk: It's for the future, probably not worth it.

mmatczuk:
Sent patchset v2

PR #526, opened 2023-11-14
Improve handling of errors from body.Close() 

Choraden:


internal/martian/proxy.go
```diff
@@ -728,7 +729,7 @@ func (p *Proxy) handle(ctx *Context, conn net.Conn, brw *bufio.ReadWriter) error
 		}
 		return errClose
 	}
-	defer req.Body.Close()
+	defer closeBody(req.Context(), req.Body, &retErr)
```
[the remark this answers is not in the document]
Choraden: Why wrong?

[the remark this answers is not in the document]
mmatczuk: Maybe it's not wrong but passing ptr to the return value is a form of hack.

Other thing is that we should return body close error only if there are no other errors.

IMO there should be `defer req.Body.Close()` and a manual body close by the end the way you handle files for instance.

internal/martian/handler.go
```diff
@@ -164,7 +164,7 @@ func (p proxyHandler) handleConnectRequest(ctx *Context, rw http.ResponseWriter,
 		res = p.errorResponse(req, cerr)
 		p.warning(res.Header, cerr)
 	}
-	defer res.Body.Close()
+	defer mustCloseBody(res.Request.Context(), res.Body)
```
mmatczuk, line 167: I believe handler does not need that because the server closes the body close and manages the connections i.e. it's out of our control.

Choraden: Of course!

internal/martian/proxy.go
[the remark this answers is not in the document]
Choraden: Got it.

e2e/tests/proxy_test.go
```diff
@@ -287,3 +291,32 @@ func TestProxyUpstream(t *testing.T) {
...
+			return nil, errors.New("only one dial is allowed")
+		}
+	})
+
+	for _, h := range []string{
+		"http://wronghost1",
+		"http://wronghost2",
+	} {
+		res := c.GET(h, func(req *http.Request) {
+			var body bytes.Buffer
```
mmatczuk, line 314: What if you try to use io.Pipe for a body or a limit reader of 10MB reading zeros?

e2e/tests/proxy_test.go
```diff
@@ -287,3 +291,32 @@ func TestProxyUpstream(t *testing.T) {
...
+			if dialed.CompareAndSwap(false, true) {
+				return d.DialContext(ctx, network, addr)
+			}
+			return nil, errors.New("only one dial is allowed")
+		}
+	})
+
+	for _, h := range []string{
+		"http://wronghost1",
+		"http://wronghost2",
```
mmatczuk, line 311: Host does not matter, you could as well have a plain loop.

e2e/tests/proxy_test.go
Choraden: Added io.Pipe with zeros to be read.

mmatczuk:
Drop the last commit it does not belong here. Also it only logs in one place.

mmatczuk:
Otherwise lgtm.

PR #527, opened 2023-11-14
WIP: martian improved error reporting

mmatczuk:
The error flow needs to be fixed, but this should target 1.2.


mmatczuk:
Work moved to #631

PR #537, opened 2023-11-16
chore(dial): simplify Dialer creation

mmatczuk:
NewDialer never fails do not return error.
Config is never read remove cfg field.


Choraden (approved):
Good.

PR #580, opened 2023-11-30
add local monitoring package and a grafana dashboard 

Choraden:
forwarder-dashboard1
forwarder-dashboard2

forwarder-dashboard3
forwarder-dashboard4
forwarder-dashboard5


mmatczuk:
Can Go routines and threads be a diagram to see how things change over time

mmatczuk:
Not exactly sure what we do now but latency should be measured for connect and upgrades separately from normal http requests.

mmatczuk:
For things that require code changes please add separate issue for 1.2

mmatczuk:
* System bytes does not have Mean and Max unlike others.
* I guess all diagrams would benefit from info especially if other ppl are to use the dashboard
* Should Go Therads be Threads?
* The threads also could have mean max
* In Response Codes you have Mean XXX req/s I guess req can be remove; also, I'm not sure what's the benefit of this information
* I think we could split Response code Rates for 2XX and everything else, this could improve resolution for the latter diagram
* Latencies seems not to go down

image

* It would be super cool if after make up it printer Grafana URL I could click to open the page

mmatczuk:
Nice I agree that adding increments for error codes is a good idea.

Overall LGTM.

docs/content/tutorials/monitoring.md
```diff
@@ -4,4 +4,67 @@ title: Monitoring
 
 # Monitoring
 
-TBD
+Monitoring your forward proxy is essential for ensuring its health, performance, and resource utilization.
```
mmatczuk, line 7: " your forward proxy "?

docs/content/tutorials/monitoring.md
```diff
@@ -4,4 +4,67 @@ title: Monitoring
...
+1. Run Dashboard with the Right Namespace:
+   - Set the `prom_namespace` variable to the value configured in your Forwarder with the `--prom-namespace` flag.
+
+Congratulations! You have successfully set up monitoring for your Forwarder using Prometheus and Grafana. Adjust the configuration parameters as needed based on your specific setup.
+
+## Try our local monitoring setup
+
+We have provided a local monitoring setup that you can use to try out the monitoring features of the Forwarder.
+All you need is a Docker to run Prometheus, Grafana and Forwarder.
+Visit [here](https://github.com/saucelabs/forwarder/blob/main/local/monitoring/README.md) for more info.
```
mmatczuk, line 70: See [local/monitoring](https://github.com/saucelabs/forwarder/blob/main/local/monitoring) for more details.

PR #596, opened 2023-12-11
log/stdlog: add log line labels

mmatczuk:
Adding stdlog.WithLabels("foo", "bar")

logger := stdlog.New(c.logConfig, stdlog.WithOnError(onError), stdlog.WithLabels("foo", "bar"))

Produces

```
2023/12/11 12:16:38.145946 [foo] [bar] [proxy] [INFO] no upstream proxy specified
2023/12/11 12:16:38.145949 [foo] [bar] [proxy] [INFO] localhost proxying mode=deny
```

### DIFF
--- a/log/stdlog/stdlog.go
+++ b/log/stdlog/stdlog.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"log"
 	"os"
+	"strings"
 
 	flog "github.com/saucelabs/forwarder/log"
 )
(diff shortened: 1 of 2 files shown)


log/stdlog/stdlog.go
```diff
@@ -59,17 +61,24 @@ type Logger struct {
...
-	sl.infoPfx = name + "[INFO] "
-	sl.debugPfx = name + "[DEBUG] "
+	sl.errorPfx = logLinePrefix(sl.labels, name, "ERROR")
+	sl.infoPfx = logLinePrefix(sl.labels, name, "INFO")
+	sl.debugPfx = logLinePrefix(sl.labels, name, "DEBUG")
 
 	return &sl
 }
 
+func logLinePrefix(labels []string, name, level string) string {
```
Choraden, line 71: It should be also called in `New`, shouldn't it?

mmatczuk: Why?

mmatczuk:
> Can we have name first?

The order follows the order of printing them. 
The labels come first so that you can cut the fixed length prefix easily.

log/stdlog/stdlog.go
Choraden: Logs won't have labels until `Named()` is called.

mmatczuk: Good point.

Choraden (approved):

PR #600, opened 2023-12-14
ci: add govulncheck-action

mmatczuk:


Choraden:
Looking at a few of the latest run actions[1] I don't really see the time benefit.

Separate actions total duration:
  - Test 1m 4s
  - E2E 7m 13s
  - Lint 23s 
 
vs previous Go action total duration from 3 random runs:
  - 7m 4s
  - 8m 2s
  - 7m 9s

If you like the jobs being separated I'm good with it, but as you can see Test and Lint jobs time contribution to Go action are negligible.

[1] https://github.com/saucelabs/forwarder/actions

mmatczuk:
> If you like the jobs being separated I'm good with it, but as you can see Test and Lint jobs time contribution to Go action are negligible.

It's just a beginning.

mmatczuk:
But I think we could join lint and test.

Choraden (approved):

PR #601, opened 2023-12-14
http_proxy: do not manually set Connection: close header

mmatczuk:
One should never set Connection header if not owning the connection. If ran in http2 server setting it causes the whole connection to be closed.


Choraden (approved):
That's right.

PR #617, opened 2023-12-21
dialvia(http): read CONNECT response by byte

### LINKED ISSUE #616: Potential data loss in dialer via HTTP proxy
Hello,

I think there is potential data loss possible in routine responsible for connection via upstream HTTP CONNECT proxy:

https://github.com/saucelabs/forwarder/blob/3bb09ce8707de386003389b0fc4e496cf11b61b0/dialvia/http.go#L110C1-L110C1

It wraps connection into a buffered reader and then it uses such buffered reader to [read upstream proxy response](https://github.com/saucelabs/forwarder/blob/3bb09ce8707de386003389b0fc4e496cf11b61b0/dialvia/http.go#L147). But buffered reader will read as much as it is available in the socket up to reader's buffer size. If some data happen to arrive right together with proxy response, reader will consume that chunk of data as well. Once response is read, that buffered reader and content in its buffer is discarded, so under certain rare conditions there is a possible data loss.

It's not an issue for protocols where client speaks first (HTTP, TLS and so on), server won't speak before client anyway. However, I think it might be an issue for protocols where servers speaks first right after client connection (SMTP, FTP, SSH, ...).

I see three potential solutions:

* Read byte by byte until proxy response is read, but still return dialed connection not wrapped.  This is how curl reads HTTP connect proxy response.
* Retain wrapping and return conn wrapped with buffered reader.
* Wrap connection into an object which feed leftovers in buffered reader first and then continue to reading from connection.

mmatczuk:
Add byteReader, a reader that reads only one byte at a time. Wrap connection in byteReader when passing to bufio.NewReaderSize() - bufio is required due to http.ReadResponse() api.

Fixes #616

### DIFF
--- a/dialvia/http.go
+++ b/dialvia/http.go
@@ -12,6 +12,7 @@ import (
 	"crypto/tls"
 	"encoding/base64"
 	"fmt"
+	"io"
 	"net"
 	"net/http"
 	"net/http/httputil"
@@ -107,7 +108,7 @@ func (d *HTTPProxyDialer) DialContextR(ctx context.Context, network, addr string
 	}
 
 	pbw := bufio.NewWriterSize(conn, 1024)
-	pbr := bufio.NewReaderSize(conn, 1024)
+	pbr := bufio.NewReaderSize(byteReader{conn}, 1)
 
 	req := http.Request{
 		Method: http.MethodConnect,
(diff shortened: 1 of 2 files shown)


Choraden (approved):
LGTM

PR #618, opened 2023-12-21
mk: add run target to simplify local running

mmatczuk:
Config is read from .forwarder.yaml file.
It's copied from example and not tracked in git.
It binds to localhost only and runs with race flag.


.forwarder.yaml.example
```diff
@@ -0,0 +1,3 @@
+address: localhost:3128
+log-level: debug
+log-http: headers
```
Choraden, line 3: api:errors?

.gitignore
```diff
@@ -51,3 +51,6 @@ tags
 
 # scratch repo to hack
 /scratch
+
+## Local config file
+.forwarder.yaml
```
Choraden, line 56: Add empty line at the end of file please.

Choraden (approved):
Thanks

PR #620, opened 2023-12-21
dialvia(http): change read buffer size to hold response line

mmatczuk:
http.ReadResponse() calls bufio.Reader.ReadLine() to read lines. When line fits we can reuse memory and prevent additional allocation and GC.

The status line should fit below 64 bytes:
HTTP/1.1 (8+1 bytes)
Status-Code (3+1 bytes)
Reason-Phrase (<=31 bytes)
CRLF (2 bytes)
Total bytes: 8 + 1 + 3 + 1 + 31 = 44

With 128 bytes we should be able to read most header lines too.


Choraden (approved):
Thanks!

PR #630, opened 2024-01-03
command/run: add default GOMEMLIMIT

mmatczuk:
Setting GOMEMLIMIT has very positive impact on the GC policy. It allows to defer GC til there is less work.
In my tests, using mainly local/browser test, the Heap usage is bellow 40MiB and the system allocated memory bellow 100MiB.


Choraden (commented):
LGTM.

We need to document it. How about we add a `tunning` section on the front page?

Choraden (approved):

PR #631, opened 2024-01-03
martian: improve error logging in response writing

mmatczuk:
This patch extracts response writing to separate method and uses isClosedConnError() for smarter error logging like we do in case if readRequest().

CloseAfrerReply is removed for easier refactoring and later function reuse.

### DIFF
--- a/internal/martian/proxy.go
+++ b/internal/martian/proxy.go
@@ -703,9 +703,9 @@ func (p *Proxy) handle(ctx *Context, conn net.Conn, brw *bufio.ReadWriter) error
 		}
 
 		if isClosedConnError(err) {
-			log.Debugf(context.TODO(), "connection closed prematurely: %v", err)
+			log.Debugf(context.TODO(), "connection closed prematurely while reading request: %v", err)
 		} else {
-			log.Errorf(context.TODO(), "failed to read request: %v", err)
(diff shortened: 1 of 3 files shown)


internal/martian/proxy.go
```diff
@@ -806,26 +806,35 @@ func (p *Proxy) handle(ctx *Context, conn net.Conn, brw *bufio.ReadWriter) error
...
 
-func (p *Proxy) writeResponse(res *http.Response, brw *bufio.ReadWriter, conn net.Conn) (closing bool) {
+func (p *Proxy) writeResponse(res *http.Response, brw *bufio.ReadWriter, conn net.Conn) error {
 	req := res.Request
 
 	if p.WriteTimeout > 0 {
 		if deadlineErr := conn.SetWriteDeadline(time.Now().Add(p.WriteTimeout)); deadlineErr != nil {
 			log.Errorf(req.Context(), "can't set write deadline: %v", deadlineErr)
 		}
+		defer conn.SetWriteDeadline(time.Time{})
```
Choraden, line 825: Nice catch!

Choraden: We should also reset the timer while reading.

mmatczuk: For reading it's ok due to tricky code.

PR #637, opened 2024-01-05
Add armed Listener with TLS handshake timeout

Choraden:
This patch adds server side TLS handshake timeout. It will protect forwarders listener and additionally a handshake in mitm. 

Due to the lack of a convenient solution upstream connect is given a constant 1m timeout to complete.

### DIFF
--- a/http_proxy.go
+++ b/http_proxy.go
@@ -258,6 +258,7 @@ func (hp *HTTPProxy) configureProxy() error {
 	hp.proxy.RequestIDHeader = hp.config.RequestIDHeader
 	hp.proxy.ConnectRequestModifier = hp.config.ConnectRequestModifier
 	hp.proxy.ConnectFunc = hp.config.ConnectFunc
+	hp.proxy.ConnectTimeout = 60 * time.Second
 	hp.proxy.WithoutWarning = true
 	hp.proxy.ErrorResponse = hp.errorResponse
 	hp.proxy.IdleTimeout = hp.config.IdleTimeout
--- a/http_proxy.go
+++ b/http_proxy.go
@@ -251,6 +251,7 @@ func (hp *HTTPProxy) configureProxy() error {
 				return hp.config.MITMDomains.Match(req.URL.Hostname())
 			}
 		}
+		hp.proxy.MITMTLSHandshakeTimeout = hp.config.TLSServerConfig.HandshakeTimeout
 	}
 
 	hp.proxy.AllowHTTP = true
--- a/http_proxy.go
+++ b/http_proxy.go
@@ -28,7 +28,6 @@ import (
 	"github.com/saucelabs/forwarder/log"
 	"github.com/saucelabs/forwarder/middleware"
 	"github.com/saucelabs/forwarder/pac"
-	"github.com/saucelabs/forwarder/ratelimit"
 	"github.com/saucelabs/forwarder/ruleset"
 )
 
@@ -109,6 +108,9 @@ func DefaultHTTPProxyConfig() *HTTPProxyConfig {
 			Protocol:          HTTPScheme,
 			Addr:              ":3128",
 			ReadHeaderTimeout: 1 * time.Minute,
+			TLSServerConfig: TLSServerConfig{
+				HandshakeTimeout: 10 * time.Second,
+			},
 		},
 		Name:            "forwarder",
 		ProxyLocalhost:  DenyProxyLocalhost,
@@ -599,27 +601,26 @@ func (hp *HTTPProxy) Run(ctx context.Context) error {
 }
 
 func (hp *HTTPProxy) listen() (net.Listener, error) {
-	listener, err := Listen("tcp", hp.config.Addr)
-	if err != nil {
-		return nil, fmt.Errorf("failed to open listener on address %s: %w", hp.config.Addr, err)
-	}
-
-	if rl, wl := int64(hp.config.ReadLimit), int64(hp.config.WriteLimit); rl > 0 || wl > 0 {
-		// Notice that the ReadLimit stands for the read limit *from* a proxy, and the WriteLimit
-		// stands for the write limit *to* a proxy, thus the ReadLimit is in fact
-		// a txBandwidth and the WriteLimit is a rxBandwidth.
-		listener = ratelimit.NewListener(listener, wl, rl)
-	}
-
 	switch hp.config.Protocol {
-	case HTTPScheme:
-		return listener, nil
-	case HTTPSScheme, HTTP2Scheme:
-		return tls.NewListener(listener, hp.tlsConfig), nil
+	case HTTPScheme, HTTPSScheme, HTTP2Scheme:
 	default:
-		listener.Close()
 		return nil, fmt.Errorf("invalid protocol %q", hp.config.Protocol)
 	}
+
+	l := Listener{
+		Address:             hp.config.Addr,
+		Log:                 hp.log,
+		TLSConfig:           hp.tlsConfig,
+		TLSHandshakeTimeout: hp.config.TLSServerConfig.HandshakeTimeout,
+		ReadLimit:           int64(hp.config.ReadLimit),
+		WriteLimit:          int64(hp.config.WriteLimit),
+	}
+
+	if err := l.Listen(); err != nil {
+		return nil, err
+	}
+
+	return &l, nil
 }
 
 // Addr returns the address the server is listening on.
--- a/internal/martian/proxy.go
+++ b/internal/martian/proxy.go
@@ -121,6 +121,9 @@ type Proxy struct {
 	// Implementations can return ErrConnectFallback to indicate that the CONNECT request should be handled by martian.
 	ConnectFunc ConnectFunc
 
+	// ConnectTimeout specifies the maximum amount of time to connect to upstream before cancelling request.
+	ConnectTimeout time.Duration
+
 	// MITMFilter specifies a function to determine whether a CONNECT request should be MITMed.
 	MITMFilter func(*http.Request) bool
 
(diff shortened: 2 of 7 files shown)


http_proxy.go
```diff
@@ -615,7 +618,9 @@ func (hp *HTTPProxy) listen() (net.Listener, error) {
 	case HTTPScheme:
 		return listener, nil
 	case HTTPSScheme, HTTP2Scheme:
-		return tls.NewListener(listener, hp.tlsConfig), nil
+		l := NewTLSListener(listener, hp.tlsConfig)
```
mmatczuk, line 622: How about we passed TLSServerConfig to NewTLSListener?

mmatczuk:
It feels like tlshandshake could be inlined given that we have the TLSListener

mmatczuk:
Some more context it the commit messages would be helpful

http_proxy.go
Choraden: I decided to leave it as it is, to have logging in the same place where we load certs. 
```go
func (hp *HTTPProxy) configureHTTPS() error {
	if hp.config.CertFile == "" && hp.config.KeyFile == "" {
		hp.log.Infof("no TLS certificate provided, using self-signed certificate")
	} else {
		hp.log.Debugf("loading TLS certificate from %s and %s", hp.config.CertFile, hp.config.KeyFile)
	}

	hp.tlsConfig = httpsTLSConfigTemplate()

	return hp.config.ConfigureTLSConfig(hp.tlsConfig)
}
```

Choraden:
> It feels like tlshandshake could be inlined given that we have the TLSListener

Listener is one thing. The other is handling handshake in mitm or SC. 
It can be later extended with proper error handling.

http_proxy.go
Choraden: Eventually, I changed it.

mmatczuk:
Now we can inline handshake in the listener.

Choraden:
I've came to a conclusion adding handshake in the listener might not be the best option. 

The logic in the martian Proxy.Serve would cause Proxy to exit on a failed handshake. 
```go
conn, err := l.Accept()
nosigpipe.IgnoreSIGPIPE(conn)
if err != nil {
	var nerr net.Error
	if ok := errors.As(err, &nerr); ok && nerr.Temporary() {
		if delay == 0 {
			delay = 5 * time.Millisecond
		} else {
			delay *= 2
		}
		if max := time.Second; delay > max {
			delay = max
		}

		log.Debugf(context.TODO(), "temporary error on accept: %v", err)
		time.Sleep(delay)
		continue
	}

	if errors.Is(err, net.ErrClosed) {
		log.Debugf(context.TODO(), "listener closed, returning")
		return err
	}

	log.Errorf(context.TODO(), "failed to accept: %v", err)
	return err
}
```

net.go
```diff
@@ -78,3 +81,85 @@ func Listen(network, address string) (net.Listener, error) {
 	// I asked about it here: https://groups.google.com/g/golang-nuts/c/Q1I7Viz9AJc
 	return defaultListenConfig().Listen(context.Background(), network, address)
 }
+
+type Listener struct {
```
mmatczuk, line 85: Let's use exported variables to configure the listener like we do with http servers for instance.

net.go
```diff
@@ -78,3 +81,85 @@ func Listen(network, address string) (net.Listener, error) {
...
+	net.Listener
+	log              log.Logger
+	tlsCfg           *tls.Config
+	handshakeTimeout time.Duration
+
+	// OnAccept is called when a new connection is successfully accepted.
+	OnAccept func(net.Conn)
```
mmatczuk, line 92: An interface with the callback would be better if we are planning to have more of them.

http_proxy.go
```diff
@@ -251,6 +251,7 @@ func (hp *HTTPProxy) configureProxy() error {
 	hp.proxy.ReadTimeout = hp.config.ReadTimeout
 	hp.proxy.ReadHeaderTimeout = hp.config.ReadHeaderTimeout
 	hp.proxy.WriteTimeout = hp.config.WriteTimeout
+	hp.proxy.MITMTLSHandshakeTimeout = hp.config.TLSServerConfig.HandshakeTimeout
```
mmatczuk, line 254: Move it to mitm setup

internal/martian/proxy.go
```diff
@@ -155,6 +155,10 @@ type Proxy struct {
 	// A zero or negative value means there will be no timeout.
 	WriteTimeout time.Duration
 
+	// MITMTLSHandshakeTimeout specifies the maximum amount of time to wait for a TLS handshake for a MITMed connection.
+	// Zero means no timeout.
+	MITMTLSHandshakeTimeout time.Duration
```
mmatczuk, line 160: Move it adjanct to MITMFilter

Choraden:
@mmatczuk I've addressed your comments

net.go
```diff
@@ -78,3 +82,96 @@ func Listen(network, address string) (net.Listener, error) {
...
+}
+
+func (l *Listener) Listen() error {
+	ll, err := Listen("tcp", l.Address)
+	if err != nil {
+		return err
+	}
+
+	if rl, wl := l.ReadLimit, l.WriteLimit; rl > 0 || wl > 0 {
+		// Notice that the ReadLimit stands for the read limit *from* a proxy, and the WriteLimit
```
mmatczuk, line 113: Let's invert and document that.

ReadLimit is for reading form the returned connections and is shared for all connections and WriteLimit is for writing to the returned connections.

net.go
```diff
@@ -78,3 +82,96 @@ func Listen(network, address string) (net.Listener, error) {
...
+			}
+			continue
+		}
+
+		return tc, nil
+	}
+}
+
+func (l *Listener) withTLS(c net.Conn) (net.Conn, error) {
+	tc := tls.Server(c, l.TLSConfig)
```
mmatczuk, line 152: use Martian/std naming conn

net.go
```diff
@@ -78,3 +82,96 @@ func Listen(network, address string) (net.Listener, error) {
...
+
+	// OnTLSHandshakeError is called after a TLS handshake errors out.
+	OnTLSHandshakeError(*tls.Conn, error)
+}
+
+type Listener struct {
+	Address          string
+	Log              log.Logger
+	TLSConfig        *tls.Config
+	HandshakeTimeout time.Duration
```
mmatczuk, line 98: TLSHandshakeTimeout

net.go
```diff
@@ -78,3 +82,96 @@ func Listen(network, address string) (net.Listener, error) {
...
+	ReadLimit        int64
+	WriteLimit       int64
+	Callbacks        ListenerCallbacks
+
+	listener net.Listener
```
mmatczuk, line 103: How about we embedded net.Listener?

mmatczuk: And remove address?

net_test.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/saucelabs/forwarder/log"
+)
+
+func TestListenerTLSHandshakeTimeout(t *testing.T) {
+	tlsCfg := new(tls.Config)
+	if err := (&TLSServerConfig{HandshakeTimeout: 100 * time.Millisecond}).ConfigureTLSConfig(tlsCfg); err != nil {
+		t.Fatal(err)
+	}
+
+	donec := make(chan struct{})
```
mmatczuk, line 26: use done as it's more popular variant.

net.go
mmatczuk: nah.

net.go
```diff
@@ -78,3 +82,96 @@ func Listen(network, address string) (net.Listener, error) {
...
+				l.Log.Errorf("Failed to close TLS connection: %v", cerr)
+			}
+			continue
+		}
+
+		return tc, nil
+	}
+}
+
+func (l *Listener) withTLS(c net.Conn) (net.Conn, error) {
```
mmatczuk, line 151: rename c to conn and return tls.Conn

net.go
Choraden: It's not the scope of this PR. Here I only moved the place where we apply it. I can make an issue to invert that.

net_test.go
```diff
@@ -0,0 +1,72 @@
...
+
+	l := Listener{
+		Address:          "[::]:0",
```
mmatczuk, line 29: Why bind to IPv6?

Tests should run on localhost as much as possible.
This way you can avoid the macOS banner to allow listening from the test process.
Please one an issue to fix that accross the board i.e. in Martian tests.

Choraden: >Why bind to IPv6?

I guess, I took it from martian tests.

internal/martian/proxy.go
```diff
@@ -512,7 +516,15 @@ func (p *Proxy) handleMITM(ctx *Context, req *http.Request, session *Session, br
...
-		if err := tlsconn.Handshake(); err != nil {
+		var hctx context.Context
+		if p.MITMTLSHandshakeTimeout > 0 {
+			var hcancel context.CancelFunc
+			hctx, hcancel = context.WithTimeout(req.Context(), p.MITMTLSHandshakeTimeout)
+			defer hcancel()
+		} else {
+			hctx = req.Context()
+		}
+		if err := tlsconn.HandshakeContext(hctx); err != nil {
```
mmatczuk, line 530: Can we have var error - less ugly code.

internal/martian/proxy.go
```diff
@@ -121,6 +121,9 @@ type Proxy struct {
 	// Implementations can return ErrConnectFallback to indicate that the CONNECT request should be handled by martian.
 	ConnectFunc ConnectFunc
 
+	// ConnectTimeout specifies the maximum amount of time to wait for a dial to complete.
```
mmatczuk, line 124: It's not dial it's connect with upstream.

Choraden:
@mmatczuk I've addressed your comments and opened two issues.

net.go
mmatczuk: ok

mmatczuk (approved):
LGTM

PR #640, opened 2024-01-09
e2e/prometheus: use latest image

Choraden:
This is to unify the image version with `local/monitoring`.


mmatczuk:
This is not what I intended but ok.

mmatczuk (approved):
LGTM

PR #646, opened 2024-01-10
Add gRPC e2e test

mmatczuk:


command/forwarder/root.go
```diff
@@ -106,12 +106,19 @@ func Command() *cobra.Command {
...
+	// Add test commands.
+	test := &cobra.Command{
+		Use:   "test",
+		Short: "Run test servers for various protocols",
+	}
+	test.AddCommand(
+		grpc.Command(),
+		httpbin.Command(),
 	)
+	cmd.AddCommand(test)
```
Choraden, line 118: nit: Test command could be moved to `/command/test/test.go`

mmatczuk: It's not as grate as it sounds. The usefulness of such command is negligible.

Choraden: ok

Choraden:
LGTM

I wish we had better solution to configure test env: `t.Setenv("HTTPS_PROXY", proxy) // set proxy for grpc.Dial` and
`$(shell docker inspect forwarder-e2e-httpbin-1 --format='{{ range .Config.Env }}{{ if eq "FORWARDER" (slice . 0 9) }} -e "HTTPBIN{{slice . 9}}"{{end}}{{end}}' 2> /dev/null)`.

mmatczuk:
I'm open to rewrite possibilities.

PR #654, opened 2024-01-17
http_proxy: support multi-addresses listening

### LINKED ISSUE #629: Support multiple ports binding
Our use case is that we run a proxy that listens on multiple ports using the same upstream.
Those ports are mentioned in [this doc](https://docs.saucelabs.com/secure-connections/sauce-connect/advanced/specifications/#supported-browsers-and-ports).

Listeners on those additional ports are optional - if the proxy fails to bind to one of the ports, it just logs an error and continues (it happens with ephemeral ports).

Would it be possible to add this functionality to Forwarder?

Choraden:
This patch adds multi-addresses listener. 

Fixes #629

### DIFF
--- a/bind/flag.go
+++ b/bind/flag.go
@@ -244,6 +244,12 @@ func HTTPServerConfig(fs *pflag.FlagSet, cfg *forwarder.HTTPServerConfig, prefix
 			"The server address to listen on. "+
 			"If the host is empty, the server will listen on all available interfaces. ")
 
+	fs.StringSliceVarP(&cfg.OptionalAddrs,
+		namePrefix+"optional-addresses", "", cfg.OptionalAddrs, "<host:port,...>"+
+			"Optional server addresses to listen on. "+
+			"The server will continue to run if the bind fails. "+
+			"Can be specified multiple times.")
+
 	if schemes == nil {
 		schemes = []forwarder.Scheme{
 			forwarder.HTTPScheme,
--- a/net.go
+++ b/net.go
@@ -10,11 +10,13 @@ import (
 	"context"
 	"crypto/tls"
 	"net"
+	"sync"
 	"syscall"
 	"time"
 
 	"github.com/saucelabs/forwarder/log"
 	"github.com/saucelabs/forwarder/ratelimit"
+	"go.uber.org/multierr"
 )
 
 type DialConfig struct {
@@ -87,12 +89,18 @@ type ListenerCallbacks interface {
 	// OnAccept is called when a new connection is successfully accepted.
 	OnAccept(net.Conn)
 
+	// OnBindError is called when a listener fails to bind to an address.
+	OnBindError(address string, err error)
+
 	// OnTLSHandshakeError is called after a TLS handshake errors out.
 	OnTLSHandshakeError(*tls.Conn, error)
 }
 
+// Listener is a multi-address listener with TLS support, rate limiting and callbacks.
+// The Listener must successfully bind on Address, but it may fail to bind on OptionalAddresses.
 type Listener struct {
 	Address             string
+	OptionalAddresses   []string
 	Log                 log.Logger
 	TLSConfig           *tls.Config
 	TLSHandshakeTimeout time.Duration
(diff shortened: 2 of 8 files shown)


bind/flag.go
```diff
@@ -93,6 +93,12 @@ func HTTPProxyConfig(fs *pflag.FlagSet, cfg *forwarder.HTTPProxyConfig, lcfg *lo
 	HTTPServerConfig(fs, &cfg.HTTPServerConfig, "", forwarder.HTTPScheme, forwarder.HTTPSScheme)
 	LogConfig(fs, lcfg)
 
+	fs.StringSliceVar(&cfg.ExtraAddresses, "extra-address", cfg.ExtraAddresses,
+		"<host:port,...>"+
+			"Additional addresses to listen on. "+
+			"The flag can be specified multiple times to add multiple addresses. "+
+			"Bind failure is not critical, the proxy will continue to run. ")
```
waggledans, line 100: can we make it configurable? In my request, bind failures shouldn't be critical (and it should be a default) but it feels like it should be more generic

Choraden: We can always make it configurable. The question is do we need it right now? If not, I'd opt for not adding it, to keep the flags simple.

waggledans: fair point. Configurability is not needed right now and this functionality is not battle-tested yet, let's see how it goes.

mmatczuk:
I'd suggest to extend the port syntax we use in the current flag

```
flag := [<host>]:<port_expr>
port_expr := <port>|<port-range>,...
port-range := <port>-<port>
```

Example:

--address localhost:50,60,70-80

Maybe the port range is too much and can be ignored.

The logic should be that we shall always be able to listen to one or more ports.

net.go
```diff
@@ -89,24 +91,64 @@ type ListenerCallbacks interface {
 
 	// OnTLSHandshakeError is called after a TLS handshake errors out.
 	OnTLSHandshakeError(*tls.Conn, error)
+
+	// OnBindError is called when a listener fails to bind to an address.
+	OnBindError(address string, err error)
```
mmatczuk, line 96: Should go before TLS things.

net.go
```diff
@@ -89,24 +91,64 @@ type ListenerCallbacks interface {
...
 }
 
+// Listener is a multi-address listener with TLS support, rate limiting and callbacks.
+// The first address in the list must bind successfully and is used as the listener address.
 type Listener struct {
-	Address          string
+	Addresses        []string
```
mmatczuk, line 102: Listener changes should be in a separate commit.

net.go
```diff
@@ -89,24 +91,64 @@ type ListenerCallbacks interface {
...
 	Log              log.Logger
 	TLSConfig        *tls.Config
 	HandshakeTimeout time.Duration
 	ReadLimit        int64
 	WriteLimit       int64
 	Callbacks        ListenerCallbacks
 
-	listener net.Listener
+	ls        []net.Listener
```
mmatczuk, line 110: Why not listeners

net.go
```diff
@@ -89,24 +91,64 @@ type ListenerCallbacks interface {
...
+	l.acceptCh = make(chan acceptResult, len(l.Addresses))
+	l.closeCh = make(chan struct{})
+
+	for i, addr := range l.Addresses {
+		ll, err := l.listen(addr)
+		if err != nil {
+			if l.Callbacks != nil {
+				l.Callbacks.OnBindError(addr, err)
+			}
+			if i == 0 {
```
mmatczuk, line 133: I think that when using more than one port they are all equal.

net.go
```diff
@@ -116,13 +158,37 @@ func (l *Listener) Listen() error {
 		ll = ratelimit.NewListener(ll, wl, rl)
 	}
 
-	l.listener = ll
-	return nil
+	return ll, nil
+}
+
+func (l *Listener) acceptLoop(ll net.Listener) {
```
mmatczuk, line 164: I'd use `errgroup` with anonymous function instead of creating this and closeCh.

net_test.go
```diff
@@ -70,3 +122,6 @@ func (m *mockListenerCallback) OnTLSHandshakeError(_ *tls.Conn, err error) {
 	}
 	m.donec <- struct{}{}
 }
+
```
mmatczuk, line 125: Order.

mmatczuk:
As for the integration, the main part should be in the listener and it should go to the HttpServer.

mmatczuk:
The `--address` and `--optional-addresses` is a usability mess it immediately raises questions like why would you ever need that.

All of that would not be a problem if we focused on file based configuration only - say we would have listener config object like Envoy. Then you could add optional field to it. We also have protocol and bunch of other stuff that make adding multiple addresses imbalance all of that.

Maybe we are solving wrong problem and this should be done on a script level?

waggledans:
> Maybe we are solving wrong problem and this should be done on a script level?

What do you have in mind? The problem is described in https://github.com/saucelabs/forwarder/issues/629

> The --address and --optional-addresses is a usability mess

I'd rather have some way to assign properties to listeners, not sure how to make it work with CLI. 

>  it immediately raises questions like why would you ever need that.

hmm... https://github.com/saucelabs/forwarder/issues/629 ?

waggledans:
> What that even means?, Why we should so easily give up on the ports, What's the purpose?

All these questions indicate that we need to improve Sauce Labs docs if it's not clear from the description.  Let's have this discussion elsewhere (we should have had it BEFORE the implementation).

> Why not use a script to launch a bunch of forwarders instead 

Absolutely not. It's 20-30 ports, I don't think so.

mmatczuk:
Let's split the work and move the port forwarding to separate feature https://github.com/saucelabs/forwarder/issues/669

PR #658, opened 2024-01-18
command/test/grpc: add plain text mode

mmatczuk:


command/test/grpc/grpc.go
```diff
@@ -70,20 +71,37 @@ func (c *command) runE(cmd *cobra.Command, _ []string) (cmdErr error) {
...
-		if err != nil {
-			return err
+		if c.plainText {
+			gs = grpc.NewServer()
+
+			var err error
+			l, err = net.Listen("tcp", c.addr)
+			if err != nil {
+				return err
+			}
```
Choraden, line 86: The listening part is duplicated in both branches. Let's move it above.

PR #673, opened 2024-02-05
chore(martian): unpack Proxy to proxyConn and move stuff around

mmatczuk:
This add `proxyConn` type to avoid passing conn and buffer as function arguments.
In addition to that move different items from proxy.go to dedicated files.


Choraden (approved):
LGTM

I guess the new files can have 2024 copyright note.

PR #683, opened 2024-02-08
martian: remove Context and Session

### LINKED ISSUE #497: martian: remove session hijacking
Martian Session can be hijacked, hijacking means getting and owning the underlying connection. 

https://github.com/saucelabs/forwarder/blob/7e70f3d02f6a9bffb2290677027b4856a9d592e1/internal/martian/context.go#L114

This feature is generally not used. We use it in proxy to implement abortIf. This implementation is problematic because every time client does something "bad" we terminate the connection. In some cases it makes sense ex. wrong basic auth credentials but in other cases it makes more harm than good.

If we removed hijacking
* the proxy code could be streamlined because we would not have to check if session was hijacked every time we ran modifiers
* abortIf could be removed, we would have normal modifiers error handling needs to be fixed in forwarder to return appropriate response for error

### LINKED ISSUE #498: Domain access denied is not an error
This is a follwup to #495. The domain access denied error log should be an info log.

### LINKED ISSUE #523: martian: replace Context and Session with context.Context
Context and Session are used to
- Hijack connection (to be removed)
- Store random attributes (can be replaced by context.Contex)
- Provide ID can be moved to context.Contex

### LINKED ISSUE #525: martian: handle ModifyRequest ModifyResponse errors
Now the errors are just logged. Either we fail to process request if modification fails or we change the API to make it explicit that we do not validate the errors.

mmatczuk:
This patchset is a significant refactoring of the martian core.
It's a continuation of work initiated in #444.

The `RequestModifier` and `ResponseModifier` errors are handled and appropriate error response is generated based on `ErrorResponse()`. Subsequently the connection hijacking functionality is removed.

Proxy connection `proxyConn` is introduced, and proxy code is restructured to more files focused around specific topics. The `proxyConn` and `proxyHandler` have similar structure for easier navigation and compatibility. The Proxy remains for configuration and generic shared items. This has been done in #673 but we expand on it here.

A `TraceID` is added and it takes the responsibility of request ID generation and duration tracking. The custom Context and Session structs are removed resulting in streamlined function interfaces. TraceID is attached to request context on creation.

Logging is streamlined, single responsibility principle is applied allowing for logger stacking. Each logger has its domain. 

Fixes #525
Fixes #523
Fixes #497
Fixes #498


log/martianlog/martianlog.go
```diff
@@ -7,10 +7,34 @@
...
+}
+
+// contextLogger is a wrapper around log.Logger that implements the martian log.Logger interface.
+type contextLogger struct {
+	log.Logger
+}
+
+var _ martianlog.Logger = contextLogger{}
+
+func (l contextLogger) Errorf(_ context.Context, format string, args ...interface{}) {
```
Choraden, line 30: Let's use `any` instead of `interface{}`. 

BTW the linter should have caught it.

Choraden:
I'm afraid, this does not fully fixes #498 as it still reports denied requests as errors in metrics. 
```
forwarder_proxy_errors_total{reason="denied"} 1
```

Let's open another one or do not resolve it here.

Choraden:
While trying new via loop detection mechanism, I found #688. 

This PR did not introduce this. It exists in the old approach as well.

PR #701, opened 2024-02-20
net: instrument dials

mmatczuk:


command/run/run.go
```diff
@@ -315,6 +315,8 @@ func Command() *cobra.Command {
 		apiServerConfig:     forwarder.DefaultHTTPServerConfig(),
 		logConfig:           log.DefaultConfig(),
 	}
+	c.httpTransportConfig.PromRegistry = c.promReg
+	c.httpProxyConfig.PromNamespace = promNs
```
alexh-sauce, line 319: Should `c.httpProxyConfig.PromNamespace = promNs` be httpTransportConfig? I see c.httpProxyConfig.PromNamespace being set twice here.

alexh-sauce (changes requested):

prom.go
```diff
@@ -0,0 +1,17 @@
...
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+package forwarder
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// promConfig is a configuration for Prometheus metrics.
+type promConfig struct {
```
Choraden, line 14: Why not exported?

Choraden: Actually, this is a nice trick!

command/run/run.go
mmatczuk: Correct.

Choraden (approved):
LGTM

PR #711, opened 2024-02-22
net(metrics): rename unspecified host to localhost

Choraden:


net_metrics.go
```diff
@@ -62,6 +62,11 @@ func addr2Host(addr string) string {
 	if err != nil {
 		return "unknown"
 	}
+
+	if host == "0.0.0.0" || host == "::" {
```
mmatczuk, line 66: Let's refactor and use `isLocalhost(req *http.Request) bool`

PR #719, opened 2024-02-23
config: Allow `@` symbols in credentials

alexh-sauce:
When using a password with an `@` in the password, we get this error string:

`only one '@' is allowed`

We should allow that symbol in passwords.


mmatczuk:
LGTM

PR #743, opened 2024-03-12
docsgen: move WriteCommand and WriteDefaultConfig to utils

Choraden:
This makes the functions importable by other projects.


mmatczuk (approved):
LGTM

PR #744, opened 2024-03-12
bind: change data URI scheme format in pac flag usage

Choraden:
'<' and '>' make a mess in sc docs.

```
-p, --pac <path or URL> (env FORWARDER_PAC)
    Proxy Auto-Configuration file to use for upstream proxy selection. It can be a local file or a URL, you can
    also use '-' to read from stdin. The data URI scheme is supported, the format is "data:base64,encoded_data".
```


Choraden:
Looks like it also wasn't working in forwarder;s docs:
Screenshot 2024-03-12 at 20 18 50

mmatczuk:
So escape it.

Choraden:
Escaped it. Notice, we also escape format in the first line '<path or URL>' as the first pair of backticks('`') is stripped in flag.UnquoteUsage(*flag.Flag).

PR #747, opened 2024-03-18
middleware(prom): add customizable label for http metrics

Choraden:
It can be used to partition the metrics by custom conditions. e.g. Label direct/tunnel mode in SC5.

PrometheusOpt was chosen to ensure compatibility with existing prom instances.

### DIFF
--- a/http_proxy.go
+++ b/http_proxy.go
@@ -93,6 +93,7 @@ type HTTPProxyConfig struct {
 	ConnectTimeout    time.Duration
 	ReadLimit         SizeSuffix
 	WriteLimit        SizeSuffix
+	PromHTTPOpts      []middleware.PrometheusOpt
 
 	// TestingHTTPHandler uses Martian's [http.Handler] implementation
 	// over [http.Server] instead of the default TCP server.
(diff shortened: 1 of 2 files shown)


middleware/prometheus.go
```diff
@@ -45,50 +54,69 @@ type Prometheus struct {
 	requestDuration  *prometheus.HistogramVec
 	requestSize      *prometheus.HistogramVec
 	responseSize     *prometheus.HistogramVec
+
+	label   string
+	labeler func(*http.Request) string
```
mmatczuk, line 59: It would be better as a type.

Choraden: ok

mmatczuk:
What is the status of this PR?

http_proxy.go
```diff
@@ -94,6 +94,7 @@ type HTTPProxyConfig struct {
 	ConnectTimeout         time.Duration
 	ReadLimit              SizeSuffix
 	WriteLimit             SizeSuffix
+	PromHTTPOpts           []middleware.PrometheusOpt
```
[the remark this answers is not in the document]
Choraden: Adding it here makes no sense. It is very specific, while PromConfig is widely used across all components.

Choraden:
Looks like it's good to go. Let's discuss it during our meeting.

PR #750, opened 2024-03-20
Document Prometheus metics  

mmatczuk:
Sample page

image


docs/content/metrics.md
```diff
@@ -0,0 +1,204 @@
...
+
+## forwarder run
+
+### `forwarder_dialer_closed_total`
+
+Number of closed connections
+
+Labels:
+  - host
+### `forwarder_dialer_dialed_total`
```
Choraden, line 15: Can we have an empty line between labels and new metric. It does not matter in the final md output, yet it just feels better.

utils/docsgen/prom.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+func WriteCommandProm(cmd *cobra.Command, promDir string) error {
+	f, err := os.Create(path.Join(promDir, "metrics.md"))
+	if err != nil {
+		return err
+	}
+
+	fmt.Fprintf(f, "---\n")
+	fmt.Fprintf(f, "title: Metrics\n")
```
Choraden, line 31: Let's also add `id: metrics`.  It's needed for sc docs.

Choraden: or `id: cmd.Name() +"metrics"`

Choraden (approved):
LGTM

PR #757, opened 2024-03-26
Use docker or podman compose instead of docker-compose

mmatczuk:
`podman compose` would preconfigure and run `docker-compose` if it can find it in the path.
It works out of the box and gives us the best of both worlds.

I also tried `podman-compose` the `docker-compose` alternative, but the performance is much worse and I failed to fix it in reasonable time.


Choraden:
LGTM

PR #761, opened 2024-03-29
e2e test in container

mmatczuk:


Makefile
```diff
@@ -74,10 +74,10 @@ update-devel-image:
 	@ln Containerfile LICENSE LICENSE.3RD_PARTY $(TMPDIR)
 ifeq ($(shell uname),Linux)
 	@CGO_ENABLED=1 GOOS=linux go build -race -o $(TMPDIR)/forwarder ./cmd/forwarder
-	@$(CONTAINER_RUNTIME) buildx build -f Containerfile --build-arg BASE_IMAGE=ubuntu:latest -t saucelabs/forwarder:$(TAG) $(TMPDIR)
+	@$(CONTAINER_RUNTIME) buildx build --network host -f Containerfile --build-arg BASE_IMAGE=ubuntu:latest -t saucelabs/forwarder:$(TAG) $(TMPDIR)
```
Choraden, line 77:
```
make update-devel-image CONTAINER_RUNTIME=docker                                                             2 ↵
unknown flag: --network
See 'docker --help'.
```
:/

Choraden:
What I really like about the current approach is the existence of `compose.yaml` directly in `/e2e`. 
I often work like this: 
1. Find the right setup.
2. Run tests with this setup to generate `compose.yaml`.
3. Manually modify the compose file e.g. add some env vars.
4. Develop/debug. 

I'd appreciate if we could somehow keep this functionality. 
Maybe we could add another target to generate compose directly with API ports exposed? 

This would also require keeping make targets like `up`, `down` to play with. BTW I think we need `make down` either way to shut down composes in `debug` mode.

Makefile
mmatczuk: That's strange it I do not see any errors in the action

https://github.com/saucelabs/forwarder/actions/runs/8522801578/job/23343764480

Choraden: It could've been my setup. If it works with pure docker then ok.

alexh-sauce (approved):
LGTM

e2e/setup/setup.go
```diff
@@ -98,7 +99,19 @@ func (r *Runner) runSetup(s *Setup) (runErr error) {
...
+				fmt.Fprintf(os.Stderr, "failed to copy compose file: %v\n", err)
+				os.Exit(1)
+			}
+
+			if err := cmd.Close(); err != nil {
+				fmt.Fprintf(os.Stderr, "failed to close compose: %v\n", err)
+				os.Exit(1)
+			}
+
+			fmt.Fprintf(os.Stdout, "To debug, run:\n\n COMPOSE_PROJECT_NAME=%s %s compose <command> ...\n\n", cmd.Runtime(), cmd.Project())
```
Choraden: Project and Runtime are swapped.

Choraden:
```
COMPOSE_PROJECT_NAME=docker compose-1048077364 compose <command> ...
```

e2e/Makefile
```diff
@@ -1,65 +1,20 @@
-include .env
+export FORWARDER_VERSION ?= devel
 
 CONTAINER_RUNTIME ?= docker
 COMPOSE := $(CONTAINER_RUNTIME) compose -f compose.yaml
```
Choraden, line 4: After removal of compose targets `COMPOSE` is not used anywhere.

e2e/run.go
```diff
@@ -69,6 +69,15 @@ func main() {
...
 						srv.Environment["FORWARDER_LOG_HTTP"] = "headers,api:errors"
 					}
+					switch srv.Name {
+					case forwarder.ProxyServiceName:
+						srv.Ports = append(srv.Ports,
+							"3128:3128",
+							"10000:10000",
+						)
+					case setup.TestServiceName:
+						srv.Command += " -test.v"
```
Choraden, line 71: This is never applied as test service is created afterwards.

Choraden:
`e2e/README.md` became outdated.

PR #765, opened 2024-04-04
Parallel e2e tests

mmatczuk:
This patchset redefines our e2e tests framework allowing to run the tests in parallel.
It also allows to run remotely with `podman-remote` as the container runtime.

Check [e2e/README.md](https://github.com/saucelabs/forwarder/blob/mmt/parallel_e2e/e2e/README.md) for details.


e2e/setup/runner.go
```diff
@@ -0,0 +1,174 @@
...
+	g, _ := errgroup.WithContext(ctx)
+	if r.Parallel > 0 {
+		g.SetLimit(r.Parallel)
+	}
+	if r.Debug {
+		g.SetLimit(1)
+	}
+
+	setups := slices.Clone(r.Setups)
+	if r.Parallel != 1 {
```
Choraden, line 46: nit: `r.Parallel != 1 && !r.Debug`

Choraden:
LGTM, the debug mode satisfies my needs. 

I'm a little surprised the e2e job takes 5m. I thought we would be able to reduce it even more, but it is still a great improvement as non parallel takes 8m on average.

mmatczuk:
Adding more parallelism does not make it run faster due to constrained resources.
We could run it in two shards odd and even to make it twice as fast.

mmatczuk:
https://github.com/saucelabs/forwarder/issues/770

PR #772, opened 2024-04-05
fix e2e DNS issues and increase parallelism level by default

mmatczuk:


e2e/run.go
```diff
@@ -31,7 +32,7 @@ var args = struct {
 	setup:    flag.String("setup", "", "Only run setups matching this regexp"),
 	run:      flag.String("run", "", "Only run tests matching this regexp"),
 	debug:    flag.Bool("debug", false, "Enables debug logs and preserves containers after running, this will run only the first matching setup"),
-	parallel: flag.Int("parallel", 1, "How many setups to run in parallel"),
+	parallel: flag.Int("parallel", runtime.NumCPU()/2+1, "How many setups to run in parallel"),
```
Choraden, line 35: Unfortunately, my Podman machine does not have access to all host's cpus. The parallelism becomes too high and some tests still fail.

Choraden: It's still DNS lookup error for services other than 'proxy'  🤔

mmatczuk:
Updated DNS wait to all services.

Choraden:
1. It looks like context cancellation leaks composes - they are not cleaned up. Let's fix it here or open an issue. 

2. Compose networks are visible for each other, thus they can't have the same name. Tests that use internal network fail when their setups happen to be executed simultaneously. Try `make -C e2e run-e2e SETUP="flag-dns"`

Other than that LGTM.

mmatczuk:
I'm aware of the issues.

Choraden (approved):

PR #787, opened 2024-04-23
http_proxy: remove ambiguity in connect headers configuration

### LINKED ISSUE #782: http_proxy: remove ambiguity in connect headers configuration
This is a small refactoring that would remove the ambiguity in Martian configuration.

Plan:
* Wrap header modifier in a layer that checks the HTTP method, for CONNECT use `proxyHeaders`, otherwise use `requestHeaders`
* Remove ConnectRequestModifier in HTTPProxyConfig, martian.Proxy and dialvia
* Add ProxyConnectHeader in dialvia and set it in Martian

Choraden:
This patch:
- renames `--proxy-header` to `--connect-header` as it was confusing
- merges connect modifiers with the classical ones to streamline its usage in proxy
- removes ambiguous ConnectRequestModifier from proxy

Fixes #782


mmatczuk:
I think that we lost the ability to set the connect headers with this patch.
The only place where `d.ProxyConnectHeader` is set is in.

```
	if id := req.Header.Get(p.RequestIDHeader); id != "" {
		h := make(http.Header)
		h.Set(p.RequestIDHeader, id)
		d.ProxyConnectHeader = h
	}
```

Please add an e2e test for this flag.

Choraden:
> I think that we lost the ability to set the connect headers with this patch.

Yes, we did and I did it on purpose. It turns out we can't reuse regular RequestHeaders as they contain basic auth and some other modifiers that should not apply to dialvia's CONNECT request. In order to do that we would need to add some extra logic, which would definitely increase complexity.
The primary reason to modify proxy connect headers was to add request-id which now we do. At this point I'm wondering if trading code simplicity and readability for having this feature is worth it. 

WDYT? @mmatczuk

mmatczuk:
AFAIMC the only mussing bit is to copy the processed headers if any to dialvia in martian.

Choraden:
1. ProxyConnectHeader in dialvia is now a full clone of headers from received CONNECT request
2. set http.Transport.GetProxyConnectHeader to 
3. added e2e tests 

I'm not sure, if we really need to support case nr 2. As you can see in tests it really takes effort to achieve it. One needs to send http request with `X-Forwarded-Proto` header explicitly set to `https`.

mmatczuk:
I agree that --connect-headers are better name can we have --proxy-headers as deprecated alias for backwards compatibility.

dialvia/http.go
```diff
@@ -125,11 +125,8 @@ func (d *HTTPProxyDialer) DialContextR(ctx context.Context, network, addr string
 		req.Header.Add("Proxy-Authorization", "Basic "+base64.StdEncoding.EncodeToString([]byte(auth)))
 	}
 
-	if cm := d.ConnectRequestModifier; cm != nil {
-		if err := cm(&req); err != nil {
-			conn.Close()
-			return nil, nil, err
-		}
+	for k, v := range d.ProxyConnectHeader {
```
mmatczuk, line 128: now you can have `maps.Copy` from stdlib

mmatczuk:
LGTM

mmatczuk (approved):

PR #789, opened 2024-04-26
dialvia: add connect timeout

### LINKED ISSUE #781: dialvia: add connect timeout
At the moment connect t/o is set as 2*dial context in main, and it's implemented as context timeout on dialer provided by dialvia in Martian. The scope of this task is moving Context t/o to dialvia and change the semantics.

Choraden:
Fixes #781

### DIFF
--- a/command/run/run.go
+++ b/command/run/run.go
@@ -198,10 +198,6 @@ func (c *command) runE(cmd *cobra.Command, _ []string) (cmdErr error) {
 		}
 	}
 
-	if 2*c.httpTransportConfig.DialTimeout > c.httpProxyConfig.ConnectTimeout {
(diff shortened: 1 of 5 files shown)


Choraden:
Thinking about the http.Transport behaviour[1] maybe we should be doing the same. This is always set a constant 60s timeout not to leak any goroutines. After all we support context cancellation. 


[1] https://cs.opensource.google/go/go/+/refs/tags/go1.22.2:src/net/http/transport.go;l=1727

mmatczuk:
This looks good to me.
```
	newCtx, cancel := context.WithTimeout(ctx, 1*time.Minute)
```

PR #791, opened 2024-04-29
e2e: cpuset per service

mmatczuk:
This makes e2e tests run in 4 minutes instead of 6.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -29,11 +29,8 @@ jobs:
       - name: Generate certificates
         run: make -C e2e/certs certs
 
-      - name: Build test helper containers
-        run: make -C e2e update-helper-images
-
-      - name: Build test container
-        run: make -C e2e update-test-image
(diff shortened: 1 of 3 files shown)


.github/workflows/e2e.yml
```diff
@@ -29,11 +29,8 @@ jobs:
       - name: Generate certificates
         run: make -C e2e/certs certs
 
-      - name: Build test helper containers
-        run: make -C e2e update-helper-images
-
-      - name: Build test container
-        run: make -C e2e update-test-image
+      - name: Build test containers
+        run: make -j4 -C e2e update-helper-images update-test-image
```
alexh-sauce, line 33: Is the 4 meant to match the # of CPUs? Could we make this `-j$(nproc)` instead?

mmatczuk: CPU is not the bottleneck here. 
We want to run all builds in parallel because they are mostly downloading image layers.

PR #796, opened 2024-05-02
perf: use goroutine pooling 

Choraden:
This patch applies ants - goroutines pooling to improve performance.

compose:
```
services:
    proxy:
        image: saucelabs/forwarder:${FORWARDER_VERSION}
        environment:
            FORWARDER_API_ADDRESS: :10000
            FORWARDER_LOG_HTTP: none
            FORWARDER_LOG_LEVEL: info
            FORWARDER_PROTOCOL: http
            GOMAXPROCS: "4"
        ports:
            - 3128:3128
            - 10000:10000
        cpus: 4
```

test: `local/browser`

allocs (old vs new):
Screenshot 2024-05-02 at 17 34 59

heap (old vs new):
Screenshot 2024-05-02 at 17 35 43

cpu (old vs new):
Screenshot 2024-05-02 at 17 36 10

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,6 +10,7 @@ require (
 	github.com/gorilla/websocket v1.5.0
 	github.com/mitchellh/go-wordwrap v1.0.1
 	github.com/mmatczuk/anyflag v0.0.0-20240105091245-c64b70bf5d41
+	github.com/panjf2000/ants/v2 v2.9.1
 	github.com/prometheus/client_golang v1.19.0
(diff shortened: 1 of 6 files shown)


mmatczuk:
Interesting. Based on the results it's not worth a merge.

Choraden:
I run another test with MultiPoolSize equal to number of CPU(4). Unfortunately, It gave the same results (<1% diff).

Choraden:
Due to fractional performance improvement it's not worth adding such a dependency.

PR #808, opened 2024-05-21
martian: fix CONNECTs content length 

Choraden:

### DIFF
--- a/internal/martian/proxy_conn.go
+++ b/internal/martian/proxy_conn.go
@@ -113,6 +113,7 @@ func (p *proxyConn) handleMITM(req *http.Request) error {
 	log.Debugf(ctx, "mitm: attempting MITM")
 
 	res := proxyutil.NewResponse(200, nil, req)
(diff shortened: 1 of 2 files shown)


internal/martian/flush.go
```diff
@@ -23,7 +23,7 @@ import (
 )
 
 func shouldFlush(res *http.Response) bool {
-	if res.Request.Method == http.MethodHead {
+	if res.Request.Method == http.MethodHead || res.Request.Method == http.MethodConnect {
```
mmatczuk, line 26: I'm not sure this should be here.

Failed connect is free to return response as any other method.

mmatczuk:
Good catch please add a test.

internal/martian/flush.go
Choraden: OK, I will drop that.

Choraden:
Tests were added.

mmatczuk (approved):
LGTM

PR #810, opened 2024-05-22
chore: simplify generating metrics docs 

mmatczuk:


command/run/run.go
```diff
@@ -431,16 +409,59 @@ func Command() *cobra.Command {
...
+		Use:                "run",
+		RunE:               c.runE,
+		DisableFlagParsing: true,
+	}
+
+	cmd.SetOut(io.Discard)
+	cmd.SetErr(io.Discard)
+
+	var stdout *os.File
+	stdout, os.Stdout = os.Stdout, os.NewFile(10, os.DevNull)
```
Choraden, line 434: What does the '10' stand for here?

Choraden (approved):

command/run/run.go
mmatczuk: file descriptor

mmatczuk:
Refactored that to log config

```
	c.logConfig = &log.Config{
		Level: log.ErrorLevel,
		File:  os.NewFile(10, os.DevNull),
	}
```

PR #812, opened 2024-05-23
martian: support server sent events without chunked encoding

Choraden:
This patch applies patternFlushWriter to support
- non-chunked SSE i.e. flush after '\n\n'
- chunked-encoding i.e. flush after '\r\n'

### DIFF
--- a/internal/martian/flush.go
+++ b/internal/martian/flush.go
@@ -18,6 +18,8 @@ package martian
 
 import (
 	"bufio"
+	"bytes"
+	"io"
 	"mime"
 	"net/http"
 )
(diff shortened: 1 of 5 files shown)


internal/martian/flush.go
```diff
@@ -53,3 +54,31 @@ func (w flushAfterChunkWriter) WriteString(s string) (n int, err error) {
 	}
 	return
 }
+
+// flushAfterSSEWriter forces a flush after each '\n\n' is written.
+type flushAfterSSEWriter struct {
```
mmatczuk, line 77: We need a flushWriter that can be used in proxy conn and handler for sse and chunked encoding.

```
type interface flusher Flush() error

type afterFlusher struct {
  w io.Writer
  f flusher
  eom [2]byte
}
```

This needs to come with a set of extensive tests.

This needs to be applied everywhere.

internal/martian/flush.go
```diff
@@ -53,3 +55,37 @@ func (w flushAfterChunkWriter) WriteString(s string) (n int, err error) {
...
+}
+
+func (w *patternFlushWriter) Write(p []byte) (n int, err error) {
+	n, err = w.w.Write(p)
+	if err != nil {
+		return
+	}
+
+	p = p[:n]
+	l := len(p)
```
mmatczuk, line 77: Isn't it the same as `n`

internal/martian/proxy_conn.go
```diff
@@ -441,9 +441,12 @@ func (p *proxyConn) writeResponse(res *http.Response) error {
 		// Add support for Server Sent Events - relay HTTP chunks and flush after each chunk.
 		// This is safe for events that are smaller than the buffer io.Copy uses (32KB).
 		// If the event is larger than the buffer, the event will be split into multiple chunks.
-		if shouldFlush(res) {
-			err = res.Write(flushAfterChunkWriter{p.brw.Writer})
-		} else {
+		switch {
+		case isTextEventStream(res):
+			err = res.Write(&patternFlushWriter{w: p.brw.Writer, f: p.brw.Writer, pattern: [2]byte{'\n', '\n'}})
```
mmatczuk, line 446: Maybe it makes sense to specify the special byte sequences as global variables.

mmatczuk, line 446: I'd prefer to have constructor.

internal/martian/flush.go
```diff
@@ -53,3 +55,37 @@ func (w flushAfterChunkWriter) WriteString(s string) (n int, err error) {
...
+type flusher interface {
+	Flush() error
+}
+
+// patternFlushWriter is an io.Writer that flushes when a pattern is detected.
+type patternFlushWriter struct {
+	w       io.Writer
+	f       flusher
+	pattern [2]byte
+	last    byte
```
mmatczuk, line 67: Can we add space before last?

internal/martian/flush.go
Choraden: Of course!

mmatczuk (approved):
LGTM

PR #814, opened 2024-05-28
martian(mitm): support CONNECT over HTTP 1.0

mmatczuk:


internal/martian/proxy_conn.go
```diff
@@ -424,9 +424,19 @@ func (p *proxyConn) writeResponse(res *http.Response) error {
 		defer p.conn.SetWriteDeadline(time.Time{})
 	}
 
-	if req.Close || p.closing() {
+	if p.closing() {
 		res.Close = true
+	} else {
+		if req.Close {
+			res.Close = true
+		}
```
Choraden, line 432: nit: the condition could be: 
```go
res.Close = req.Close
```

Choraden (approved):

internal/martian/proxy_conn.go
Choraden: nvm

PR #834, opened 2024-06-06
WIP io parallel copy

mmatczuk:

### DIFF
--- a/internal/martian/copy.go
+++ b/internal/martian/copy.go
@@ -43,22 +43,38 @@ var copyBufPool = sync.Pool{
 	},
 }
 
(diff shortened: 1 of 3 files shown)


mmatczuk:
I tested this patch with Forwarder and sc build, both locally and with dedicated Linux box 4 CPU Intel in local network.
I see no stable improvement that would justify the change.
In the test against gnet hello world, leading to this work, the new implementation (both local and over sc tunnel) does not change anything. In the test I'm able to get 20MiB/s throughput from a single core running gnet while forwarder and sc peak at 8,5MiB. 

Testing against httpbin does not really work due to high variation in latencies stddev ranging to 70-100% both with wrk and vegeta.
In vegeta this can be flattened out by using HDR but one can note how stable gnet latencies are.

PR #838, opened 2024-06-07
log: change level type to string

### LINKED ISSUE #832: log: error level is not recognised as default by anyflag
Error level evaluates to 0:
```go
const (
	ErrorLevel Level = iota
	InfoLevel
	DebugLevel
)
```
That makes anyflag ignore this value as default:
```go
func (v *Value[T]) String() string {
	if v.redact != nil {
		return v.redact(*v.value)
	}

	if v.value == nil || reflect.ValueOf(*v.value).IsZero() {
		return ""
	}
	return fmt.Sprint(*v.value)
}
```
As a result, default log level is not printed in help:
```
    --log-level <error|info|debug> (env FORWARDER_LOG_LEVEL)
        Log level.
```

Choraden:
That would eliminate the zero value issue with flags.

```
    --log-level <error|info|debug> (default error) (env FORWARDER_LOG_LEVEL)
        Log level.
```

Fixes #832

### DIFF
--- a/log/level.go
+++ b/log/level.go
@@ -6,14 +6,14 @@
 
 package log
 
-type Level int32
+type Level string
 
 const (
-	ErrorLevel Level = iota
-	InfoLevel
-	DebugLevel
+	ErrorLevel Level = "error"
+	InfoLevel  Level = "info"
+	DebugLevel Level = "debug"
 )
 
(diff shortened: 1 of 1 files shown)


mmatczuk:
LGTM

PR #847, opened 2024-06-18
log: revert log level to int

Choraden:
The previous idea was wrong. String type messes up the order of log levels.
We will shift it.


mmatczuk:
LGTM

PR #848, opened 2024-06-24
log: reopen disk files on SIGHUP

mmatczuk:


alexh-sauce (approved):

command/pac/server/server.go
```diff
@@ -38,6 +38,10 @@ func (c *command) runE(cmd *cobra.Command, _ []string) (cmdErr error) {
 	}
 	logger := stdlog.New(c.logConfig)
 
+	defer func() {
+		logger.Close()
+	}()
```
Choraden, line 43: Just `defer logger.Close()`?

mmatczuk: Actually maybe we should be printing the error to stderr

Choraden: 👍 👍

log/rotate.go
```diff
@@ -0,0 +1,67 @@
...
+
+func (w *RotatableFile) reopenOnSIGHUP() {
+	ch := make(chan os.Signal, 1)
+
+	signal.Notify(ch, syscall.SIGHUP)
+	for range ch {
+		if err := w.Reopen(); err != nil {
+			fmt.Fprintf(os.Stderr, "failed to rotate log file: %v\n", err)
+		}
+	}
```
Choraden, line 65: I'm not sure if this goroutine ends. What closes the channel?

Choraden: How about we close it in RotatableFile.Close()?

log/rotate.go
```diff
@@ -0,0 +1,79 @@
...
+	return w.file().Write(p)
+}
+
+var closeSignal = syscall.Signal(-1)
+
+func (w *RotatableFile) Close() error {
+	select {
+	case w.ch <- closeSignal:
+	default:
+	}
```
Choraden, line 46: I understand, it is to protect consecutive Close calls, but what If the channel is occupied with SIGHUPs and you only call close once. It's a rare case, but not impossible.

mmatczuk: This protects against blocking on Close() when Close is called multiple times after to Goroutine exited.

Note there is a race condition between Close and Reopen (and write) using atomics.
We could solve it by locking but then all logging would go under RW lock read path.
Given the use case I do not think it's a right approach.

There is not much we can do here tbh given this design - that's why there was no cancelation in the 1st place - it's a virtual problem.

Some of the common aids do not work here
* Closing the channel causes panics in signal and Subsequent calls to Close
* Using separate channel has the same problem
* Increasing channel buffer size does not work as you can have multiple signals queued in theory

Maybe you can suggest something that I'm missing.

Choraden: How about
1. The Close would need to call signal.Stop on the channel, so that It can safely close it. Unfortunately, it tears apart Notify from Stop.
2. use sync.Once to send closeSignal

PR #863, opened 2024-07-15
utils/compose: initial support for data containers

mmatczuk:
Add support for data sharing using data containers.

### DIFF
--- a/utils/compose/builder.go
+++ b/utils/compose/builder.go
@@ -39,6 +39,10 @@ func (b *Builder) AddService(sb ServiceBuilder) *Builder {
 	s := sb.Service()
 
 	for i, v := range s.Volumes {
+		nv, _, ok := strings.Cut(v, ":")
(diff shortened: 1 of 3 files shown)


utils/compose/builder.go
```diff
@@ -39,6 +39,10 @@ func (b *Builder) AddService(sb ServiceBuilder) *Builder {
 	s := sb.Service()
 
 	for i, v := range s.Volumes {
+		nv, _, ok := strings.Cut(v, ":")
+		if ok && b.c.Volumes[nv] != nil {
+			continue
+		}
```
Choraden, line 45: I don't see it. `Compose.AddVolume` saves whole string, but here you cut it, why?

mmatczuk: This is to support named data container like in the [following example](https://docs.docker.com/compose/compose-file/07-volumes/#example).

Choraden (approved):

PR #864, opened 2024-07-16
utils/cobrautil: FlagsDescriber change field ShowNotChanged to ShowChangedOnly

mmatczuk:
Enable displaying hidden flags that are set.


Choraden (approved):

mmatczuk:
I think this does not solve it fully. I'd write a test.

PR #866, opened 2024-07-24
Fix content length in CONNECT requests and responses

mmatczuk:
Make sure that we correctly read and send Content-Length header value for CONNECT requests.

### DIFF
--- a/internal/martian/proxy_connect.go
+++ b/internal/martian/proxy_connect.go
@@ -30,6 +30,19 @@ import (
 	"github.com/saucelabs/forwarder/internal/martian/proxyutil"
 )
 
+func fixConnectReqContentLength(req *http.Request) {
+	if req.Method != http.MethodConnect {
+		return
+	}
+
+	if req.Header.Get("Content-Length") != "" {
+		log.Infof(req.Context(), "CONNECT request with Content-Length: %s, ignoring content length",
+			req.Header.Get("Content-Length"))
+	}
+
+	req.ContentLength = -1
+}
+
 // ErrConnectFallback is returned by a ConnectFunc to indicate
 // that the CONNECT request should be handled by martian.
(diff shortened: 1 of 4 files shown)


mmatczuk:
I need to add test here.

internal/martian/proxy_connect.go
```diff
@@ -122,5 +121,19 @@ func (p *Proxy) connectSOCKS5(req *http.Request, proxyURL *url.URL) (*http.Respo
 		return nil, nil, err
 	}
 
-	return proxyutil.NewResponse(200, http.NoBody, req), conn, nil
+	return newConnectResponse(req), conn, nil
+}
+
+func newConnectResponse(req *http.Request) *http.Response {
+	return &http.Response{
+		StatusCode: http.StatusOK,
```
Choraden, line 129: Please set `Status` too. The response log lacks it, you can see it in commit message.
```
sc-1         | HTTP/2.0
```
vs 
```
sc-server-1  | HTTP/1.1 200 OK
```

mmatczuk: Done and donner

Choraden: Wrong 🤔. That's how it's done in proxyutil 
```
Status:     fmt.Sprintf("%d %s", code, http.StatusText(code)),
```

mmatczuk:
> I wonder if we still need that, since we set the ContentLength in read. I understand this is just a sanity check if the ContentLength was modified in Martian req/res modifiers?

I'd leave that to be on the safe side, we can add some logging.

internal/martian/proxy_connect.go
mmatczuk: That is the best way of doing it, fixed.

Choraden:
Sure, let's log it.

Choraden (approved):
LGTM

PR #867, opened 2024-07-25
chore(utils/httpbin): make delay return 200 on contex cancel explicit

mmatczuk:

### DIFF
--- a/utils/httpbin/httpbin.go
+++ b/utils/httpbin/httpbin.go
@@ -69,8 +69,9 @@ func delayHandler(w http.ResponseWriter, r *http.Request) {
 	case <-r.Context().Done():
 		t.Stop()
 	case <-t.C:
-		w.WriteHeader(http.StatusOK)
 	}
+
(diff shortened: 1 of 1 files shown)


utils/httpbin/httpbin.go
```diff
@@ -69,8 +69,9 @@ func delayHandler(w http.ResponseWriter, r *http.Request) {
 	case <-r.Context().Done():
 		t.Stop()
```
Choraden, line 70: Since we are here, shouldn't we also defer the timer stop?

Choraden: Don't mention it. It's a timer, not ticker.

Choraden (approved):

utils/httpbin/httpbin.go
mmatczuk: Sure, I'll add it.

mmatczuk:
Improved commit messages, added deferred timer Stop() call.

PR #870, opened 2024-07-26
martian: fix half-close propagation downstream when using http2 handler

mmatczuk:


internal/martian/errors.go
```diff
@@ -45,7 +45,8 @@ func isClosedConnError(err error) bool {
 	if errors.Is(err, io.EOF) ||
 		errors.Is(err, io.ErrUnexpectedEOF) ||
 		errors.Is(err, syscall.ECONNABORTED) ||
-		errors.Is(err, syscall.ECONNRESET) {
+		errors.Is(err, syscall.ECONNRESET) ||
+		errors.Is(err, errClosedBody) {
```
Choraden, line 49: Yes! I've seen a lot of these recently.

internal/martian/unsafe.go
```diff
@@ -0,0 +1,10 @@
+// Copyright 2022-2024 Sauce Labs Inc., all rights reserved.
+
+package martian
+
+import (
+	_ "unsafe"
+)
+
+//go:linkname errClosedBody golang.org/x/net/http2.errClosedBody
+var errClosedBody error
```
Choraden, line 10: Isn't that file name too general?

internal/martian/proxy_handler.go
```diff
@@ -337,27 +337,36 @@ func (p proxyHandler) handleRequest(rw http.ResponseWriter, req *http.Request) {
...
-	rc *http.ResponseController
+type h2Writer struct {
+	w     io.Writer
+	flush func() error
+	close func() error
 }
 
-func (w writeFlusher) Write(p []byte) (n int, err error) {
-	n, err = w.rw.Write(p)
+func makeH2Writer(rw http.ResponseWriter, rc *http.ResponseController, req *http.Request) *h2Writer {
```
Choraden, line 346: Return a pointer? Previous flusher was a copy, and the methods receive a copy too, so I'm double checking if that's on purpose.

internal/martian/unsafe.go
mmatczuk: Initially I called it h2Err... but reverted to that due to linter

internal/martian/proxy_handler.go
mmatczuk: Nah, that's too fast generation...

internal/martian/unsafe.go
Choraden: Yeah, I was thinking of some `h2error.go`

.golangci.yml
```diff
@@ -141,6 +141,10 @@ issues:
         - golint
       source: 'import _ "embed"'
 
+    - linters:
+        - revive
+      source: '_ "unsafe"'
```
Choraden, line 146: grep for already existing comments

internal/martian/proxy_handler.go
mmatczuk: Fixed

internal/martian/unsafe.go
mmatczuk: go:linkname moved to errors.go and renamed to `h2ErrClosedBody`.

.golangci.yml
mmatczuk: Dropped that commit.

Choraden (approved):
LGTM

PR #874, opened 2024-07-31
Fix reading api-address from config file in container

### LINKED ISSUE #873: docker: use unix socket for health checks
Currently health checks add additional assumptions that prevent api-address config option working in docker container if passed in a config file.

Expose api server, regardless of --api-address flag, over unix socket if running in a container and use that for health checks.

mmatczuk:
This patchset implements checking readiness over a Unix socket.
Unnecessary API servers in e2e tests are removed.
The ready command implementation is extended and improved for better reuse.   

Fixes #873

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -2,6 +2,8 @@ ARG BASE_IMAGE=gcr.io/distroless/static:nonroot
 
 FROM ${BASE_IMAGE}
 
+ENV PLATFORM=container
+
 COPY LICENSE /licenses/
 COPY LICENSE.3RD_PARTY /licenses/
 COPY forwarder /usr/bin
--- a/Containerfile
+++ b/Containerfile
@@ -8,5 +8,4 @@ COPY forwarder /usr/bin
 ENTRYPOINT ["/usr/bin/forwarder"]
 CMD ["run"]
 
-ENV FORWARDER_API_ADDRESS="localhost:10000"
 HEALTHCHECK --interval=1s --timeout=3s --retries=10 CMD ["/usr/bin/forwarder", "ready"]
--- a/utils/httpx/httpx.go
+++ b/utils/httpx/httpx.go
@@ -1,0 +1,49 @@
+// Copyright 2022-2024 Sauce Labs Inc., all rights reserved.
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+package httpx
+
+import (
+	"context"
+	"errors"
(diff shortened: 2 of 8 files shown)


Containerfile
```diff
@@ -2,6 +2,8 @@ ARG BASE_IMAGE=gcr.io/distroless/static:nonroot
 
 FROM ${BASE_IMAGE}
 
+ENV PLATFORM=container
```
Choraden, line 5: How about `FORWARDER_PLATFORM`? Something to make it difficult to overwrite.

utils/httpx/httpx.go
```diff
@@ -0,0 +1,44 @@
...
+			return ctx
+		},
+	}
+
+	go func() {
+		<-ctx.Done()
+		s.Close()
+	}()
+
+	return s.Serve(l)
```
Choraden, line 43: It returns an error on server close 
```
^C2024/08/01 08:12:27.518332 [ERROR] fatal error exiting: http: Server closed
exit status 1
```

utils/httpx/httpx.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"context"
+	"fmt"
+	"net"
+	"net/http"
+	"os"
+	"time"
+)
+
+func ServeUnixSocket(ctx context.Context, h http.Handler, socketPath string) error {
```
Choraden, line 18: May we add some logging here?

Containerfile
mmatczuk: I don't know the `FORWARDER_PLATFORM` does not make much sense to me. 

We have two situations:
* The `PLATFORM` env var get overwritten when running container - for that we could log warning in the readiness check.
* Our container behavior kicking in in another setting due to presence of the env var - here I think we are good since we must match both key and value (container)  

Let me add some more logging instead.

utils/httpx/httpx.go
mmatczuk: Any logging should go in handler.

Containerfile
mmatczuk: I considered adding sth like this

```
	if p := os.Getenv("PLATFORM"); p != "" {
		if  p == "container" {
			c.apiAddr = "unix:" + cfg.APIUnixSocket
		} else {
			fmt.Fprintf(os.Stderr, "unexpected PLATFORM environment variable: %q\n", p)
		}
	}
```

But I guess it's not really useful.

Choraden: Ugly^

Choraden: Maybe I'm exaggerating, why would somebody add PLATFORM variable to forwarder container.

Choraden (approved):
LGTM

Issue #895, opened 2024-09-04
net: stop using go:linkname


mmatczuk:
Go 1.23 prevents us from accessing unexpected parts of standard library.

At the moment we use

1. `config.go://go:linkname isDomainName net.isDomainName`
1. `utils/osdns/resolvconf.go://go:linkname resolvConf net.resolvConf`
1. `utils/osdns/resolvconf.go://go:linkname getSystemDNSConfig net.getSystemDNSConfig`
1. `resolver.go://go:linkname lookupStaticHost net.lookupStaticHost`

We need to figure out what to do case by case.

mmatczuk:
Caused by #894

PR #899, opened 2024-09-05
linkname: remove usage of net.lookupStaticHost

mmatczuk:
Relates to #895


alexh-sauce (approved):

http_proxy_test.go
```diff
@@ -109,3 +109,37 @@ func TestNopDialer(t *testing.T) {
 		t.Fatalf("expected %v, got %v", nopDialerErr, err)
 	}
 }
+
+func TestIsLocalhost(t *testing.T) {
+	cfg := DefaultHTTPProxyConfig()
+	p, err := NewHTTPProxy(cfg, nil, nil, nil, stdlog.Default())
```
Choraden, line 115: May we have a test for a handler mode too (`newHTTPProxy`)?

Choraden: That suite should work without reading hosts file.

Choraden:
Other than that - LGTM

http_proxy_test.go
mmatczuk: We can do that in a separate issue. It uses convention in this file.

PR #903, opened 2024-09-09
Refactor osdns to avoid using custom linker flags in common situations

mmatczuk:
Rename osdns to dnshack and require dnshack build tag to use it.
This allows to use standard, unmodified Go build chain at the expense of not using custom DNS parameters in dev builds.


mmatczuk:
Fixes #895

budziam (approved):

dns_dnshack.go
```diff
@@ -0,0 +1,17 @@
...
+
+package forwarder
+
+import (
+	"github.com/saucelabs/forwarder/utils/dnshack"
+)
+
+func (c *DNSConfig) Apply() error {
+	return dnshack.Configure(c.Servers, c.Timeout, c.RoundRobin)
+}
```
Choraden, line 17: Add endl

Choraden:
Don't we want to add the tag to every file in the dnshack package?

mmatczuk:
At first I did not want to add it at all here. But it turns out I needed to add them to avoid go test with linker flag. I'd add the build tag everywhere for consistency.

mmatczuk:
Added new line and build tags to all files.

Choraden:
LGTM

PR #905, opened 2024-09-10
Add --connect-to flag that allows to change dial target

### LINKED ISSUE #315: Add --connect-to flag like in curl
In curl

```
       --connect-to <HOST1:PORT1:HOST2:PORT2>

              For a request to the given HOST1:PORT1 pair, connect to
              HOST2:PORT2 instead.  This option is suitable to direct requests
              at a specific server, e.g. at a specific cluster node in a
              cluster of servers. This option is only used to establish the
              network connection. It does NOT affect the hostname/port that is
              used for TLS/SSL (e.g. SNI, certificate verification) or for the
              application protocols. "HOST1" and "PORT1" may be the empty
              string, meaning "any host/port". "HOST2" and "PORT2" may also be
              the empty string, meaning "use the request's original
              host/port".

              A "host" specified to this option is compared as a string, so it
              needs to match the name used in request URL. It can be either
              numerical such as "127.0.0.1" or the full host name such as
              "example.org".

              --connect-to can be used several times in a command line

              Example:
               curl --connect-to example.com:443:example.net:8443 https://example.com

              See also --resolve and -H, --header. Added in 7.49.0.
```

This would allow to add more flexibility / hackability, it's much easier and more flexible than DNS hacking and `/etc/hosts`.

mmatczuk:
Fixes #315

### DIFF
--- a/bind/flag.go
+++ b/bind/flag.go
@@ -238,6 +238,17 @@ func DialConfig(fs *pflag.FlagSet, cfg *forwarder.DialConfig, prefix string) {
 			"With or without a timeout, the operating system may impose its own earlier timeout. For instance, TCP timeouts are often around 3 minutes. ")
 }
 
+func ConnectTo(fs *pflag.FlagSet, cfg *[]forwarder.HostPortPair) {
+	fs.Var(anyflag.NewSliceValue[forwarder.HostPortPair](*cfg, cfg, forwarder.ParseHostPortPair),
+		"connect-to", "<HOST1:PORT1:HOST2:PORT2>,..."+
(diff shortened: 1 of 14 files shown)


Choraden:
So a `HostPort.Validate` works differently than `HostPortUser.Validate`, which embeds `HostPort` 🤔 . 
I guess, that's ok, since HostPortUser is used only in credentials (allows wildcards `*`).

Choraden:
How about adding an e2e test?

mmatczuk:
HostPortUser.Validate() now calls `HostPort.Validate()`.
Added e2e test.

Choraden (approved):

PR #909, opened 2024-09-13
e2e: enable goleak check

### LINKED ISSUE #516: e2e: enable goleak check
Find a way to assert goleak check in e2e tests in CI

Choraden:
This patch extends compose with functionality to check exit codes of services. It improves tests quality as it enforces that all services exit with code 0. That allows to correctly assert goleak errors. 

Fixes #516

### DIFF
--- a/e2e/setup/runner.go
+++ b/e2e/setup/runner.go
@@ -189,7 +189,20 @@ func (r *Runner) runSetup(s *Setup) (runErr error) {
 	}
 
 	// Run the test service.
-	return cmd.Up("--force-recreate", "--exit-code-from", TestServiceName, TestServiceName)
+	if err := cmd.Up("--force-recreate", "--exit-code-from", TestServiceName, TestServiceName); err != nil {
+		return err
+	}
+
+	if !r.Debug {
+		if err := cmd.Stop(); err != nil {
+			return fmt.Errorf("compose stop: %w", err)
+		}
+		if err := cmd.ExitedSuccessfully(r.services(s)); err != nil {
+			return fmt.Errorf("services failed to exit: %w", err)
+		}
+	}
+
+	return nil
 }
 
 func (r *Runner) services(s *Setup) []string {
--- a/e2e/setups.go
+++ b/e2e/setups.go
@@ -65,7 +65,8 @@ func SetupDefaults(l *setupList) {
 							WithProtocol(httpbinScheme)).
 					AddService(
 						forwarder.ProxyService().
-							WithProtocol(proxyScheme)).
+							WithProtocol(proxyScheme).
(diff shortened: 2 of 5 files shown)


mmatczuk:
@Choraden what's the status?

Choraden:
I've encountered goleak errors as if the proxy would not close all the goroutines gracefully, but rather a forced shutdown happens. I had no time to investigate it further.

e2e/setups.go
```diff
@@ -66,7 +66,8 @@ func SetupDefaults(l *setupList) {
 							WithProtocol(httpbinScheme)).
 					AddService(
 						forwarder.ProxyService().
-							WithProtocol(proxyScheme)).
+							WithProtocol(proxyScheme).
+							WithGoleak()).
```
mmatczuk, line 70: I wonder, why not enable it by default?

utils/compose/cmd.go
```diff
@@ -106,6 +107,41 @@ func (c *Command) Stop(args ...string) error {
 	return c.quietRun(c.cmd("stop", args))
 }
 
+func (c *Command) ExitedSuccessfully(services []string) error {
```
mmatczuk, line 110: It would be more useful to export serviceExitCode and add plain functions to do assertions about it.

e2e/setup/runner.go
```diff
@@ -189,7 +189,20 @@ func (r *Runner) runSetup(s *Setup) (runErr error) {
...
-	return cmd.Up("--force-recreate", "--exit-code-from", TestServiceName, TestServiceName)
+	if err := cmd.Up("--force-recreate", "--exit-code-from", TestServiceName, TestServiceName); err != nil {
+		return err
+	}
+
+	if !r.Debug {
+		if err := cmd.Stop(); err != nil {
+			return fmt.Errorf("compose stop: %w", err)
+		}
+		if err := cmd.ExitedSuccessfully(r.services(s)); err != nil {
```
mmatczuk, line 200: Don't we want to check exit status of all services?

mmatczuk:
This is very good progress.

e2e/setups.go
Choraden: Right. I enabled goleak in a proxy service by default.

utils/compose/cmd.go
Choraden: compose.Command now exports ServiceExitCode.

e2e/setup/runner.go
Choraden: It does check exit code of all services.

Choraden:
@mmatczuk I've addressed your comments.

mmatczuk (approved):
LGTM

PR #913, opened 2024-09-17
log/stdlog: allow to pass options when creating a sublogger

budziam:
I want to be able to modify the sublogger level. It should be possible to pass custom options when creating a sublogger.

### DIFF
--- a/log/stdlog/stdlog_test.go
+++ b/log/stdlog/stdlog_test.go
@@ -1,0 +1,20 @@
+// Copyright 2022-2024 Sauce Labs Inc., all rights reserved.
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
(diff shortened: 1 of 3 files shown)


log/stdlog/stdlog_test.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"testing"
+
+
+	flog "github.com/saucelabs/forwarder/log"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestLogger_NamedAllowsToPassCustomLevel(t *testing.T) {
```
mmatczuk, line 17: We do not use _ here

mmatczuk:
Ok.

mmatczuk (approved):

PR #915, opened 2024-09-17
Add support for proxy protocol in listener

### LINKED ISSUE #912: Support proxy protocol for downstream connection
Allow reading connection information using PROXY protocol.

mmatczuk:
Added server options

```
    --proxy-protocol-enabled <value> (default false) (env FORWARDER_PROXY_PROTOCOL_ENABLED)
        The PROXY protocol is used to correctly read the client's IP address. When enabled the proxy will expect the
        client to send the PROXY protocol header before the actual request. It is still possible to connect to the
        proxy without the PROXY protocol header when this flag is enabled.

    --proxy-protocol-read-header-timeout <duration> (default 10s) (env FORWARDER_PROXY_PROTOCOL_READ_HEADER_TIMEOUT)
        The amount of time to wait for PROXY protocol header. Zero means no limit.
```

Fixes #912

### DIFF
--- a/bind/flag.go
+++ b/bind/flag.go
@@ -201,6 +201,17 @@ func MITMDomains(fs *pflag.FlagSet, cfg *[]ruleset.RegexpListItem) {
 			"Prefix domains with '-' to exclude requests to certain domains from being MITMed.")
 }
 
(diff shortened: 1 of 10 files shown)


mmatczuk:
Closing this as we want something with better listener design.

PR #916, opened 2024-09-18
Add support for proxy protocol in martian

mmatczuk:
This is version of #915 that implements reading proxy headers in martian.
This has many benefits unfortunately this approach is wrong so I'm sending it for information purposes only.

> The PROXY protocol will take precedence over TLS when establishing a network connection. Here’s why:
> The PROXY protocol is designed to pass client connection information (like the real IP address and port) from a load balancer or proxy to a backend server. It works at the transport layer and is placed before any higher-layer protocol like TLS.

This means we need implementation in forwarder.Listener I'd send another patch.


Choraden:
I'm beginning to think we made a mistake with putting tls to the listener.

mmatczuk:
> I'm beginning to think we made a mistake with putting tls to the listener.

Why?

PR #919, opened 2024-09-20
proxyproto: PROXY protocol net.Conn and net.Listener impl

mmatczuk:
Use github.com/pires/go-proxyproto header parsing and provide better (safer) net.Conn and net.Listener implementations.

### DIFF
--- a/proxyproto/proxyproto.go
+++ b/proxyproto/proxyproto.go
@@ -1,0 +1,177 @@
+// Copyright 2022-2024 Sauce Labs Inc., all rights reserved.
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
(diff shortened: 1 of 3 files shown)


proxyproto/proxyproto.go
```diff
@@ -0,0 +1,177 @@
...
+		br := bufio.NewReaderSize(byteReader{c.Conn}, bufSize)
+
+		var r result
+		r.header, r.err = proxyproto.Read(br)
+
+		if r.err == nil && br.Buffered() > 0 {
+			panic("proxy protocol header read: unexpected data after header")
+		}
+
+		resCh <- r
```
Choraden, line 137: It may leak.

mmatczuk: Let's make buffered chanel.

PR #920, opened 2024-09-23
Support for PROXY protocol v1 and v2

mmatczuk:


net.go
```diff
@@ -163,6 +175,13 @@ func (l *Listener) Listen() error {
 		return err
 	}
 
+	if l.ProxyProtocolConfig != nil {
+		ll = &proxyproto.Listener{
+			Listener:          ll,
+			ReadHeaderTimeout: l.ProxyProtocolConfig.ReadHeaderTimeout,
+		}
```
Choraden, line 182: Use constructor? It's from different package.

Choraden: On the other hand, this is handy

Choraden:
LGTM

PR #939, opened 2024-10-14
martian(test): implement testHelper pattern

Choraden:
When I was trying to change martian behaviour that needed some tests to be adjusted, I realised they are in a poor condition. This patch introduces a testHelper to add some structure and reuse similar code blocks. That way it reduces around 150 LoC. It also leaves the tests in better shape for further refactors.


mmatczuk:
Sweet.

PR #941, opened 2024-10-14
martian: move TLS handshake from forwarder.Listener

### LINKED ISSUE #917: martian: move TLS handshake from Listener to matrtian
The forwarder.Listener preforms explicit TLS handshake if TLS is configured.
While it's beneficial to martian to the handshake before trying to read request,
it may lead to slowness in accepting new connections as the handshake is performed in proxy main loop.

Move TLS handshake out of Listener and move it to martian and make it in separate go routine.

Choraden:
Listener is not the right place to do the handshake. It may block main accept loop.
Martian Proxy will explicitly do the handshake if accepted connection is tls.Conn.

TODO: 
- [ ] add a test

Fixes #917


mmatczuk:
The question is what do we do with the tls handshake metric?

Choraden:
I rebased and added a test. 

>The question is what do we do with the tls handshake metric?

That approach resigns from having a tls error metrics in forwarder. We don't have any metrics directly in martian and I see no point in adding it only for that.
If we want to have it, we need to implement conn wrapper that would do the handshake.

internal/martian/proxy_conn.go
```diff
@@ -40,19 +40,28 @@ type proxyConn struct {
...
 	if tconn, ok := conn.(*tls.Conn); ok {
+		ctx := context.Background()
+		if p.TLSHandshakeTimeout > 0 {
+			var cancel context.CancelFunc
+			ctx, cancel = context.WithTimeout(context.Background(), p.TLSHandshakeTimeout)
+			defer cancel()
+		}
+		if err := tconn.HandshakeContext(ctx); err != nil {
+			return nil, fmt.Errorf("failed to do TLS handshake: %w", err)
+		}
```
mmatczuk, line 59: Perhaps this should be extracted to method so that `newProxyConn` does not return error.
This would give us the opportunity to provide better errors.

Choraden: A separate method to do the handshake has been added.

internal/martian/proxy_conn.go
```diff
@@ -41,18 +41,28 @@ type proxyConn struct {
...
 		Proxy: p,
 		brw:   bufio.NewReadWriter(bufio.NewReader(conn), bufio.NewWriter(conn)),
 		conn:  conn,
 	}
+}
 
-	if tconn, ok := conn.(*tls.Conn); ok {
-		v.secure = true
-		v.cs = tconn.ConnectionState()
+func (p *proxyConn) tlsHandshake() error {
```
mmatczuk, line 51: I'd call that maybeHandshakeTLS

internal/martian/proxy_conn.go
```diff
@@ -41,18 +41,28 @@ type proxyConn struct {
...
+	if tconn, ok := p.conn.(*tls.Conn); ok {
```
mmatczuk, line 52: Return early.

mmatczuk (approved):

mmatczuk:
LGTM

PR #943, opened 2024-10-16
conntrack: add connection tracker wrapper based on connfu

mmatczuk:
Use builder pattern to return implementation with only required features. Use connfu to handle CloseWrite() and ReadFrom() propagation.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/saucelabs/forwarder
 
-go 1.23
+go 1.23.1
 
 require (
 	github.com/dop251/goja v0.0.0-20231027120936-b396bb4c349d
@@ -11,6 +11,7 @@ require (
 	github.com/kevinburke/hostsfile v0.0.0-20220522040509-e5e984885321
 	github.com/mitchellh/go-wordwrap v1.0.1
 	github.com/mmatczuk/anyflag v0.0.0-20240709090339-eb9e24cd1b44
+	github.com/mmatczuk/connfu v0.0.0-20241015064402-db8989f89d8c
 	github.com/prometheus/client_golang v1.20.5
 	github.com/prometheus/client_model v0.6.1
 	github.com/prometheus/common v0.60.0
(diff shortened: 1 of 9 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/saucelabs/forwarder
 
-go 1.23
+go 1.23.1
```
Choraden, line 3: GO_VERSION=1.23.2

tracked/tracked.go
```diff
@@ -0,0 +1,142 @@
...
+
+func (c *closeConn) Close() error {
+	return c.l.Close()
+}
+
+type closeListener struct {
+	close   func() error
+	once    sync.Once
+	onClose func()
+}
```
Choraden, line 57: How about merging closeConn and closeListener?

tracked/tracked.go
```diff
@@ -0,0 +1,142 @@
...
+	// TrackTraffic enables counting of bytes read and written by the connection.
+	// Use Rx and Tx to get the number of bytes read and written.
+	TrackTraffic bool
+
+	// OnClose is called after the underlying connection is closed and before the Close method returns.
+	// OnClose is called at most once.
+	OnClose func()
+}
+
+func (b Builder) Build(c net.Conn) (net.Conn, *ConnObserver) {
```
Choraden, line 100: I'm not sure if this signature is comfortable enough. Even if you wrap many connections, you only track traffic once. This is most of the time *ConnObserver is nil.

Choraden: Maybe it could provide a set of functions that build certain types.

Choraden:
Although `tracked` seems cool, we could use some fancy name to emphasise it's a real net.Conn on steroids. 
Some proposals:
- connx
- turobconn
- hyperconn
- connboost

mmatczuk:
Traditionally the name is `conntrac` not to get confused with conntrac in Linux.

go.mod
mmatczuk: Ran go mod tidy on 1.23.2 same result.

Choraden: It's the minimal go version that can build this package.

Choraden:
@mmatczuk This looks good!

1. I wonder if we want to add assertion in martian that the splice optimisation is to take place.
2. I would love to see some numbers on how it improves performance.

PR #949, opened 2024-11-07
http proxy: allow multiple listeners

mmatczuk:
Add HTTPProxyConfig.ExtraListeners and refactor code to work with multiple listeners.

### DIFF
--- a/http_proxy.go
+++ b/http_proxy.go
@@ -17,7 +17,6 @@ import (
 	"net/http"
 	"net/url"
 	"slices"
-	"sync"
 	"time"
 
 	"github.com/saucelabs/forwarder/hostsfile"
@@ -28,6 +27,8 @@ import (
 	"github.com/saucelabs/forwarder/log"
 	"github.com/saucelabs/forwarder/middleware"
 	"github.com/saucelabs/forwarder/pac"
+	"go.uber.org/multierr"
+	"golang.org/x/sync/errgroup"
 )
 
 type ProxyLocalhostMode string
@@ -80,6 +81,7 @@ var ErrConnectFallback = martian.ErrConnectFallback
 
 type HTTPProxyConfig struct {
 	HTTPServerConfig
+	ExtraListeners    []NamedListenerConfig
 	Name              string
 	MITM              *MITMConfig
 	MITMDomains       Matcher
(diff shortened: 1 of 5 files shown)


http_proxy.go
```diff
@@ -500,79 +504,116 @@ func (hp *HTTPProxy) handler() http.Handler {
...
+			hp.log.Errorf("failed to shutdown server error=%s", err)
 		}
-		return srvErr
+		return ctx.Err()
+	})
+	for i := range hp.listeners {
+		l := hp.listeners[i]
+		g.Go(func() error {
+			return srv.Serve(l)
+		})
```
Choraden, line 535:
```
^C2024/11/07 14:41:01.894475 [proxy] [INFO] closing down proxy
2024/11/07 14:41:01.894515 [proxy] [INFO] waiting for connections to close
2024/11/07 14:41:01.894522 [proxy] [INFO] all connections closed
2024/11/07 14:41:01.894584 [ERROR] fatal error exiting: http: Server closed
exit status 1
```
Closing a proxy in a handler mode results in a fatal error.

http_proxy.go
```diff
@@ -500,79 +504,116 @@ func (hp *HTTPProxy) handler() http.Handler {
...
-	l := Listener{
-		ListenerConfig: hp.config.ListenerConfig,
-		TLSConfig:      hp.tlsConfig,
-		PromConfig: PromConfig{
-			PromNamespace: hp.config.PromNamespace,
-			PromRegistry:  hp.config.PromRegistry,
-		},
+	prototype := Listener{
+		TLSConfig: hp.tlsConfig,
+		metrics:   newListenerMetrics(hp.config.PromRegistry, hp.config.PromNamespace),
```
Choraden, line 569: How about adding a `name` to the ListenerConfig and using it in metrics? That way we could have them more detailed e.g. which listener accepts the most.

Choraden:
LGTM, personally I would go with a wrapper over a list of listeners that returns any accepted connection from the underlying slice, but this is even better!

Left a proposition on adding "named" listeners.

mmatczuk:
MultiListener is complicated its main disadvantage is that you accept a connection and then send it to worker. It's not guaranteed there is a worker available. It could be done but the standard interfaces are not well suited for that.

http_proxy.go
mmatczuk: Sure it's a good idea.

mmatczuk:
Having considered it I'd add static address label so that we can distinguish them

mmatczuk:
@Choraden PTAL

mmatczuk:
Squashed some commits and rebased.

Choraden (approved):

Issue #938, opened 2024-10-10
Websockets do not connect on first attempt on Firefox


pulsejet:
## Expected Behavior

Websockets work

## Actual Behavior

The first connection to websocket on loading a page fails, but the second one succeeds. Refresh the page and this happens again.

## Steps to Reproduce the Problem

  1. Set up forwarder+proxy in Firefox
  1. Go to https://websocketking.com/
  1. Click "Connect". It fails.
  1. Click "Connect" again. It connects now.

## Specifications

  - Version: 1.36.2
  - Platform: Windows Firefox 125

![{A884EBF0-7BE3-4482-AC31-EE47BECE2D14}](https://github.com/user-attachments/assets/a088df3d-1795-45d6-af52-2fdb244fa792)
![{7569F0A7-D266-4C1E-B044-C439FD8B109B}](https://github.com/user-attachments/assets/20b7b117-e241-4215-9ab0-3dda3953913f)
![{A659E9BA-BD10-4D06-B9CD-8000CBB30911}](https://github.com/user-attachments/assets/dd654153-7ea6-4250-8903-0e4d16b4f972)

mmatczuk:
Nice find. It work with MITM, it does not work without MITM.

mmatczuk:
What happens is that the FF is interpreting the connect response as server response.

PR #956, opened 2024-11-12
Fix connect response that prevented proper working of websockets in FireFox

mmatczuk:
Fixes https://github.com/saucelabs/forwarder/issues/938

### DIFF
--- a/internal/martian/proxy_handler.go
+++ b/internal/martian/proxy_handler.go
@@ -171,16 +171,13 @@ func (p proxyHandler) tunnel(name string, rw http.ResponseWriter, req *http.Requ
 		}
 		defer conn.Close()
 
-		if _, err := brw.WriteString(connectResponse); err != nil {
-			err := fmt.Errorf("got error while writing response back to client: %w", err)
-			p.traceWroteResponse(res, err)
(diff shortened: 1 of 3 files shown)


internal/martian/proxy_handler.go
```diff
@@ -171,16 +171,13 @@ func (p proxyHandler) tunnel(name string, rw http.ResponseWriter, req *http.Requ
...
-		if err := brw.Flush(); err != nil {
-			err := fmt.Errorf("got error while flushing response back to client: %w", err)
-			p.traceWroteResponse(res, err)
-			return err
+		pc := proxyConn{
+			Proxy: p.Proxy,
+			brw:   brw,
+			conn:  conn,
 		}
+		pc.writeResponse(res)
```
Choraden, line 179: Fine, but it starts to look odd. Maybe there is some better way to share that code then 🤔.

Choraden (approved):
LGTM with minor suggestions

internal/martian/proxy_handler.go
mmatczuk: It looks ok to me tbh

PR #963, opened 2024-11-20
Do not close client connection after returning HTTP status code 304 

mmatczuk:


internal/martian/proxy_test.go
```diff
@@ -485,6 +485,77 @@ func TestIntegrationHTTP101SwitchingProtocols(t *testing.T) {
 	}
 }
 
+func TestIntegrationHTTP304NotModified(t *testing.T) {
```
budziam, line 488: thought this test was designed to reproduce the error we encountered with a closed connection when returning a 304 response code. However, it seems the test is already passing on the main branch. Shouldn't it be failing there?

internal/martian/flush.go
```diff
@@ -36,21 +36,19 @@ func shouldChunk(res *http.Response) bool {
...
 		return false
 	}
 
-	return true
+	return !isHeadOnlyRfc7230(res)
+}
+
+// isHeadOnlyRfc returns true if the response should be a head-only response according to spec.
+// Please read 3.3.2 and 3.3.3 of RFC 7230 for more details https://datatracker.ietf.org/doc/html/rfc7230#section-3.3.2.
+func isHeadOnlyRfc7230(res *http.Response) bool {
```
Choraden, line 48:
```
 3.3.3.1
Any response to a HEAD request and any response with a 1xx
(Informational), 204 (No Content), or 304 (Not Modified) status
code is always terminated by the first empty line after the
header fields, regardless of the header fields present in the
message, and thus cannot contain a message body
```
Let's change the link to section 3.3.3 only.

Choraden: and what about 1xx responses:
```go
if res.StatusCode < 200 {
 	return false
 }
 ```
 ?

File: internal/martian/proxy_test.go [hunk 3]
```diff
@@ -485,6 +485,77 @@ func TestIntegrationHTTP101SwitchingProtocols(t *testing.T) {
...
+	res, err := http.ReadResponse(bufio.NewReader(conn), nil)
+	if err != nil {
+		t.Fatalf("http.ReadResponse(): got %v, want no error", err)
+	}
+	defer res.Body.Close()
+
+	if got, want := res.StatusCode, 304; got != want {
+		t.Fatalf("res.StatusCode: got %d, want %d", got, want)
+	}
+	if _, err := conn.Read(make([]byte, 0)); err != nil {
```
Choraden, line 554: Let's read at least a byte.

mmatczuk: There is no byte, it would block, changing to nil.

mmatczuk:
Fixed implementations renamed some functions

budziam (approved):

internal/martian/proxy_test.go
```diff
@@ -485,6 +485,77 @@ func TestIntegrationHTTP101SwitchingProtocols(t *testing.T) {
...
+	res, err := http.ReadResponse(bufio.NewReader(conn), nil)
+	if err != nil {
+		t.Fatalf("http.ReadResponse(): got %v, want no error", err)
+	}
+	defer res.Body.Close()
+
+	if got, want := res.StatusCode, 304; got != want {
+		t.Fatalf("res.StatusCode: got %d, want %d", got, want)
+	}
+	if _, err := conn.Read(nil); err != nil {
```
Choraden, line 554: What's the behavior of Read(nil)?

mmatczuk: In Go, when you call conn.Read(nil) on a net.Conn object, the behavior is as follows:
	1.	No data is read: Since the provided buffer is nil, there is no space for any data to be stored. However, the function call is still valid.
	2.	Behavior depends on the state of the connection:
	•	If there is no data available and the connection is still open, conn.Read(nil) will block until some data arrives or the connection is closed.
	•	If the connection is closed gracefully by the peer, conn.Read(nil) will return 0, io.EOF.
	•	If the connection is closed due to an error, the method will return 0, <some error>.
	3.	No actual read operation is performed: Since no buffer is provided, the method does not attempt to copy data. It’s effectively a no-op for data transfer, but it still triggers the underlying I/O and checks the connection’s state.

This behavior can be useful if you want to check the connection state or block until data is available without actually reading the data at that moment. However, it’s rarely used in practice because it doesn’t perform any meaningful work with the data.

PR #965, opened 2024-11-26
net: use KeepAliveConfig from standard net package

Choraden:
With new go version we were given more intelligent KeepAliveConfig. Using it in forwarder.Listener will allow users for custom TCP keepalive settings. By default we will remain with OS-specific parameters.


mmatczuk:
Nice to see my hacks go away

mmatczuk (approved):
LGTM

PR #966, opened 2024-11-26
martian(mitm): fix connect error response propagation

### LINKED ISSUE #510: mitm: handle upstream proxy error responses for CONNECT requests
When HTTP client fails HTTPS CONNECT request, it does not return the original response, but an error.

mmatczuk:
Fixes #510

### DIFF
--- a/internal/martian/proxy_connect.go
+++ b/internal/martian/proxy_connect.go
@@ -29,6 +29,7 @@ import (
 
 	"github.com/saucelabs/forwarder/dialvia"
 	"github.com/saucelabs/forwarder/internal/martian/log"
+	"github.com/saucelabs/forwarder/internal/martian/proxyutil"
 )
 
 func fixConnectReqContentLength(req *http.Request) {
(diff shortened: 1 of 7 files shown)


internal/martian/proxy_connect.go
```diff
@@ -201,3 +202,34 @@ func shouldTerminateTLS(req *http.Request) bool {
...
+	return e.res
+}
+
+func OnProxyConnectResponse(_ context.Context, _ *url.URL, req *http.Request, connectRes *http.Response) error {
+	if connectRes.StatusCode/100 == 2 {
+		return nil
+	}
+
+	// Body cannot be read from the CONNECT response due to use of closed network connection.
+	res := proxyutil.NewResponse(connectRes.StatusCode, http.NoBody, req)
```
Choraden, line 224: How about calling body.Close() as a best effort approach?

Choraden: N/m. I can see the conn is closed when returned error is not nil.

Choraden (approved):
LGTM

PR #967, opened 2024-11-27
On shutdown try to drain connections for 30s and close them

### LINKED ISSUE #960: systemctl restart and systemctl stop seems to hang on ubuntu 24.04 if websockets are in use
## Expected Behavior

Executing `sudo systemctl restart forwarder` should stop the deamon and disconnect any clients immediately and restart even if websocket connections are active.

## Actual Behavior

Executing it will hang as long as all websocket clients are active.  We get the same behavior with or `sudo systemctl stop forwarder`.
After all websocket clients are closed, forwarder restarts/stops.

Note: the vast majority of sites using websockets have auto-reconnect handlers in place because websocket streams are so unreliable.

## Steps to Reproduce the Problem

  1. Run forwarder on a proxy server
  2. Setup proxy in web browser to use the proxy server
  3. Connect to a website using websockets
  4. Try to execute `sudo systemctl restart forwarder` (or `sudo systemctl stop forwarder`) on the proxy server
  5. Watch nothing happen :-)

## Specifications

  - Version: v1.3.2
  - Platform: Ubuntu 24.04
  - Subsystem:

mmatczuk:
Shutdown tries to drain connections for 30s or until terminate signal is received, then they are closed. 

Fixes #960

### DIFF
--- a/http_proxy.go
+++ b/http_proxy.go
@@ -16,6 +16,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"runtime"
 	"slices"
 	"strings"
 	"time"
@@ -188,6 +189,7 @@ func NewHTTPProxy(cfg *HTTPProxyConfig, pr PACResolver, cm *CredentialsMatcher, 
 		return nil, err
 	}
 	hp.listeners = ll
+	runtime.SetFinalizer(hp, (*HTTPProxy).closeListeners)
 
 	for _, l := range hp.listeners {
 		hp.log.Infof("PROXY server listen address=%s protocol=%s", l.Addr(), hp.config.Protocol)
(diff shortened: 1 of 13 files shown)


http_proxy.go
```diff
@@ -628,22 +649,11 @@ func (hp *HTTPProxy) Addr() (addrs []string, ok bool) {
...
-
-	// Shutdown the proxy to stop serving requests.
-	if e := hp.proxy.Shutdown(context.Background()); e != nil {
-		err = multierr.Append(err, e)
-	}
-
-	if tr, ok := hp.transport.(*http.Transport); ok {
-		tr.CloseIdleConnections()
-	}
-
```
Choraden, line 634: So now there is a `Close` that closes listeners, but user still needs to cancel ctx. 
Let's hide `Close` method and use in it Run. This way all workflow will be in `Run`.

mmatczuk: After implementing it I think the Close method gives users possibility for cleanup without relying on GC. It's actually kind of sad that the variable is not freed when it goes out of scope i.e. test function ends but that's how it is.

I'm reverting the last commit.

Choraden: ok

PR #968, opened 2024-11-28
Fix local/browser and add mk targets for profiling

mmatczuk:


loadgen/topgun.py
```diff
@@ -1,11 +1,10 @@
-import threading
```
alexh-sauce, line 1: This shouldn't need to be moved. Python imports are traditionally sorted alphabetically.

mmatczuk: It was auto formatted and now imports are in alphabetical order.

Choraden (approved):

loadgen/topgun.py
alexh-sauce: Weird, was I reading that diff backwards, or maybe I can't do alphabetical order in my head!? Sorry for the noise.

PR #972, opened 2024-12-03
metrics: connections use active instead of closed 

### LINKED ISSUE #945: net: listener add active connections metric
Now active connections can be calculated by subtracting closed from accepted. It would be easier if we had separate metric.

mmatczuk:
Fixes #945


Choraden (approved):
LGTM

Issue #676, opened 2024-02-06
e2e: add wrk test


mmatczuk:
Add a test target that allows to test the proxy using wrk.

Sample command

```
wrk http://127.0.0.1:3128 -d 10 -t 1 -c 200
```

PR #974, opened 2024-12-04
martian/header: optimized ViaModifier implementation

mmatczuk:

### DIFF
--- a/internal/martian/header/via_modifier.go
+++ b/internal/martian/header/via_modifier.go
@@ -20,78 +20,25 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"regexp"
 	"strings"
 
 	"github.com/saucelabs/forwarder/internal/martian"
 )
 
-var whitespace = regexp.MustCompile("[\t ]+")
+const (
+	h20Prefix = "2.0"
+	h11Prefix = "1.1"
+	h10Prefix = "1.0"
+)
 
 // ViaModifier is a header modifier that checks for proxy redirect loops.
 type ViaModifier struct {
-	requestedBy string
-	boundary    string
+	tag string
 }
 
 // NewViaModifier returns a new Via modifier.
 func NewViaModifier(requestedBy string) *ViaModifier {
-	return &ViaModifier{
-		requestedBy: requestedBy,
-		boundary:    randomBoundary(),
-	}
-}
-
-// ModifyRequest sets the Via header and provides loop-detection. If Via is
-// already present, it will be appended to the existing value. If a loop is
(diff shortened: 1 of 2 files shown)


internal/martian/header/via_modifier.go
```diff
@@ -49,42 +55,36 @@ func NewViaModifier(requestedBy string) *ViaModifier {
...
+		l += len(via) + 2
+	}
+
+	var sb strings.Builder
+	sb.Grow(l)
 
-	if v := req.Header.Get("Via"); v != "" {
-		if m.hasLoop(v) {
+	if via != "" {
+		if strings.Contains(via, m.requestedBy+"-"+m.boundary) {
```
Choraden, line 69: How about precomputing `m.requestedBy+"-"+m.boundary` instead of concatenating on every modify?

Choraden:
LGTM.

I feel that optimisations are correct, but it would be useful to have #676.

internal/martian/header/via_modifier.go
mmatczuk: Sure.

mmatczuk:
@Choraden check the new implementation

mmatczuk:
Here we could have a microbenchmark

Choraden:
> Here we could have a microbenchmark

That's true.

mmatczuk:
```
 michal@SL-2094:martian/header (mmt/has_loop2)$ benchstat old.txt new.txt
goos: darwin
goarch: arm64
pkg: github.com/saucelabs/forwarder/internal/martian/header
cpu: Apple M1 Pro
            │    old.txt    │               new.txt               │
            │    sec/op     │    sec/op     vs base               │
ViaModifier   2237.0n ± ∞ ¹   152.8n ± ∞ ¹  -93.17% (p=0.029 n=4)
¹ need >= 6 samples for confidence interval at level 0.95

            │   old.txt    │              new.txt               │
            │     B/op     │    B/op      vs base               │
ViaModifier   1440.0 ± ∞ ¹   160.0 ± ∞ ¹  -88.89% (p=0.029 n=4)
¹ need >= 6 samples for confidence interval at level 0.95

            │   old.txt    │              new.txt               │
            │  allocs/op   │  allocs/op   vs base               │
ViaModifier   31.000 ± ∞ ¹   4.000 ± ∞ ¹  -87.10% (p=0.029 n=4)
¹ need >= 6 samples for confidence interval at level 0.95
```

mmatczuk:
It was indeed slow

Choraden (approved):
💪

PR #981, opened 2024-12-06
net: dial retries

### LINKED ISSUE #706: net: dial retries
In case dial fails in an upstream Forwarder it's ineffective.
We should add dial retry with tiny exponential backoff.

mmatczuk:
Fixes #706

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -23,6 +23,7 @@ linters:
     - gomnd
     - gomoddirectives
     - inamedparam
+    - intrange
     - ireturn
     - mnd
     - nlreturn
--- a/net.go
+++ b/net.go
@@ -57,6 +57,11 @@ func defaultKeepAliveConfig() net.KeepAliveConfig {
 	}
 }
 
+type DialRetryConfig struct {
+	Attempts int
+	Backoff  time.Duration
+}
+
 type DialConfig struct {
 	// DialTimeout is the maximum amount of time a dial will wait for
 	// connect to complete.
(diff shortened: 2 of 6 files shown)


.golangci.yml
```diff
@@ -23,6 +23,7 @@ linters:
     - gomnd
     - gomoddirectives
     - inamedparam
+    - intrange
```
Choraden, line 26: Why not use range?

net.go
```diff
@@ -155,6 +171,37 @@ func (d *Dialer) DialContext(ctx context.Context, network, address string) (net.
...
+
+		conn, err := dial(ctx, network, address)
+		if err == nil {
+			return conn, nil
+		}
+		lastErr = err
+
+		if conn != nil {
+			conn.Close()
+		}
```
Choraden, line 200: if err != nil, conn is nil. That's the rule we were following before this patch.

Choraden: Maybe it can stay. Better safe than sorry.

net.go
```diff
@@ -79,6 +69,18 @@ func DefaultDialConfig() *DialConfig {
 	return &DialConfig{
 		DialTimeout:     25 * time.Second,
 		KeepAliveConfig: defaultKeepAliveConfig(),
+		Retry:           3,
+		RetryBackoff:    1 * time.Second,
+	}
```
Choraden: This does not belong here.

net.go
mmatczuk: That's the rule overall but there is nothing that governs that rule.

Choraden (approved):

PR #1004, opened 2025-01-27
martian: cancel request context on downstream disconnection

### LINKED ISSUE #800: martian: cancel request context when connection is closed
For Martian TCP listener implementation support request context cancelation when downstream connection dies.

Choraden:
This patch introduces a background read mechanism for martian proxy conn. Background read affects simple roundTrip path. Once the request body is consumed in roundTrip, background read will begin, canceling conn context in an error occurs. This idea was inspired by http.Server. 

This partially fixes #800. The other patch in progress.

### DIFF
--- a/http_proxy_errors.go
+++ b/http_proxy_errors.go
@@ -8,6 +8,7 @@ package forwarder
 
 import (
 	"bytes"
+	"context"
 	"crypto/tls"
 	"crypto/x509"
 	"errors"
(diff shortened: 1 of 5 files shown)


Choraden:
The requirement of using `bufio.Reader` in `http.ReadRequest` makes this problem much harder. To implement it properly, we would need to drift away from http package. I'll follow up with other solutions.

Issue #1013, opened 2025-02-12
martian: trace wrote response called multiple times on the same response


Choraden:
This messes up metrics.

PR #1014, opened 2025-02-12
martian: add graceful close timeout for bicopy

Choraden:
This will prevent the proxy from leaking connections with defective servers.

Given limited resources, this is the best we can do for #800 right now.

That fix cannot be tested due to https://github.com/saucelabs/forwarder/issues/1013. Enable the test again once that is fixed.
For now I tested it manually.

### DIFF
--- a/http_proxy_errors.go
+++ b/http_proxy_errors.go
@@ -8,6 +8,7 @@ package forwarder
 
 import (
 	"bytes"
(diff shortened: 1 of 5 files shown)


Choraden:
@budziam Changed the log level to info.

budziam (approved):

PR #1053, opened 2025-06-03
martian: fix trace wrote response is called once per request

Choraden:
traceWroteResponse would be called multiple times when handling CONNECT and protocol upgrade requests.
This patch assures there is exactly one call to wroteResponse per one call to readRequest.


markamsauce (approved):
LGTM

PR #1061, opened 2025-06-16
martian: fix trace wrote 

Choraden:
`traceWroteResponse` wasn't invoked or was wrongly invoked in some places, even after the last patch to it.
It is difficult to test all the flows and the best indicator I have right now is to collect metrics from some instances and compare results. 
This tracer design is not good and it is error prone, however let's just fix it like that for now.


markamsauce (approved):
LGTM

PR #1067, opened 2025-06-30
add formats for structured logging 

Choraden:
This patch adds two logging formats for structured logger.
Text format is more human-readable, while json is targeted at production  workloads for easier log analysis.


markamsauce (approved):
Let's merge it.

PR #1071, opened 2025-07-01
log/slog: rename log name attribute from "name" to "module"

Choraden:
It's not entirely clear what the name attribute refers to.
In my opinion, log would be a better fit—it clearly indicates that it's related to the application's logging module.


markamsauce (approved):
Looks good.

PR #1111, opened 2026-02-10
support renaming heades to non-canonicalized case

mrogala-sauce:
Support header renaming to use non-canonicalized names.

See changed docs for more explanation.


markamsauce (approved):
LGTM

PR #1113, opened 2026-08-04
[SC-5424] Sanitize malformed query with unicode replacements

sarthaksrivastava-prog:
Following this change, the forwarder will forward requests with bare % in the URL by escaping to %25, to sanitize malformed queries and redirect them towards the sanitized URLs.

### DIFF
--- a/internal/martian/proxy_conn.go
+++ b/internal/martian/proxy_conn.go
@@ -106,11 +106,11 @@ func (p *proxyConn) readRequest() (*http.Request, error) {
 		log.Error(context.TODO(), "can't set read header deadline", "error", deadlineErr)
 	}
 
-	br, err := sanitizeRequestLine(p.brw.Reader, p.conn)
-	if err != nil {
-		return nil, err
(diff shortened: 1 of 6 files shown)


internal/martian/proxy_conn.go
```diff
@@ -106,6 +106,12 @@ func (p *proxyConn) readRequest() (*http.Request, error) {
 		log.Error(context.TODO(), "can't set read header deadline", "error", deadlineErr)
 	}
 
+	br, err := sanitizeRequestLine(p.brw.Reader, p.conn)
+	if err != nil {
```
mrogala-sauce: I would suggest to only log an error and do not propagate it - if sanitizeRequestLine fails just do not replace the Reader. This is not critical path and we will not break customer connections if something goes wrong with sanitization or we hit unknown bug

sarthaksrivastava-prog: Corrected. On sanitize failure we now log.Error and keep the original reader (p.brw.Reader is not swapped), so the request continues through http.ReadRequest unchanged

mrogala-sauce (approved):
LGTM

